PR #1, opened 2019-12-05
Initial import

plorenz:


ziti/config/config.go
```diff
@@ -0,0 +1,52 @@
...
+	return &Config{
+		ZtAPI:    ztApi,
+		ID:       idConfig,
+	}
+}
+
+func NewFromFile(confFile string) (*Config, error) {
+	conf, err := ioutil.ReadFile(confFile)
+	if err != nil {
+		return nil, errors.New(fmt.Sprintf("Config file (%s) is not found ", confFile))
```
dovholuknf: this isn't a log but i could see it getting logged elsewhere. should errors follow the lowercase letters a the start rule?

ziti/config/config.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	c := Config{}
+	err = json.Unmarshal(conf, &c)
+	if err != nil {
+		return nil, errors.New(fmt.Sprintf("Failed to load ZT configuration (%s): %v", confFile, err))
```
dovholuknf: same lower/capital case comment - i won't make others but there might be more to resolve.

ziti/config/token.go
```diff
@@ -0,0 +1,117 @@
...
+		delete(mapClaims, "StandardClaims")
+
+		if stdMap, ok := std.(map[string]interface{}); ok {
+			for k, v := range stdMap {
+				if !isZeroValue(v) {
+					mapClaims[k] = v
+				}
+			}
+		} else {
+			return nil, errors.New("could not converted standard claims section to map")
```
dovholuknf: "could not converted" - typo?

ziti/config/config.go
michaelquigley: we also shouldn't be using `ZT`

ziti/edge/conn_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+package edge
+
+import (
+	"testing"
+)
+
+func Test_edgeChannel_Host(t *testing.T) {
+
+}
```
michaelquigley: Do we need to keep this around if it doesn't do anything?

ziti/config/config.go
michaelquigley: Can also use `fmt.Errorf` instead of `errors.New(fmt.Sprintf(...`

ziti/config/config.go
plorenz: fixed

ziti/config/config.go
plorenz: fixed. Switched to using errors.Errorf where appropriate and fixed casing where I found it.

ziti/config/token.go
plorenz: fixed

ziti/edge/conn_test.go
plorenz: fixed

PR #6, opened 2019-12-13
Update to latest foundation

plorenz:


example/simple-server/simple-server.go
```diff
@@ -55,6 +55,7 @@ func plain(listenAddr string) {
 func withZiti(service string) {
 	listener, err := ziti.NewContext().Listen(service)
 	if err != nil {
+		fmt.Printf("Error binding service %+v\n", err)
```
dovholuknf: is the \n really needed? shouldn't we strive for one line?

michaelquigley: You need `\n` with `Printf`. `Printf` doesn't add a newline, `Println` does.

plorenz: It's because it's not a log stmt. Without the \n there's no new line. It's an example program, so I think either fmt or log is ok here.

PR #14, opened 2020-01-07
Update for session name changes

plorenz:


example/simple-server/simple-server.go
```diff
@@ -18,10 +18,14 @@ package main
...
 	"fmt"
-	"github.com/netfoundry/ziti-sdk-golang/ziti"
 	"net"
 	"net/http"
 	"os"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/michaelquigley/pfxlog"
+	"github.com/netfoundry/ziti-sdk-golang/ziti"
```
michaelquigley: It looks like the formatting reverted back to the opposite of what we've been using everywhere else. We've traiditionally been formatting these to be in one block, in alphabetical order.

plorenz: Fixed. I had to recreate my workspace when we broke ziti up and lost my settings. I couldn't remember if we were using goland import formatting or goimports. I tried goimports, but that was apparently the wrong choice :)

PR #19, opened 2020-01-23
Allow setting which config types to request. Remove dns from services.

plorenz:


ziti/config/config.go
```diff
@@ -24,8 +24,9 @@ import (
 )
 
 type Config struct {
-	ZtAPI string                  `json:"ztAPI"`
-	ID    identity.IdentityConfig `json:"id"`
+	ZtAPI       string                  `json:"ztAPI"`
+	ID          identity.IdentityConfig `json:"id"`
+	ConfigTypes []string                `json:"configTypes"`
```
ekoby: Does this mean that types are stored in Identity file?

plorenz: They can be, but for the tunneler I'm just hardcoding them after the config is loaded. I think in general that hardcoding them makes sense, b/c what is supported is likely to be determined with the code, so having to push it to a config file and keep that in sync is just more work for more potential error.

PR #20, opened 2020-01-28
on windows the erorr that returns is a path error but not 'IsNotExist'

dovholuknf:
on windows it return an error stating the name is incorrect not that the file doesn't exist. the stat will be nil.

### DIFF
--- a/ziti/enroll/enroll.go
+++ b/ziti/enroll/enroll.go
@@ -116,7 +116,7 @@ func Enroll(enFlags EnrollmentFlags) (*config.Config, error) {
 	if strings.TrimSpace(enFlags.KeyFile) != "" {
 		stat, err := os.Stat(enFlags.KeyFile)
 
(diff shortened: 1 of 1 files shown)


ziti/enroll/enroll.go
```diff
@@ -116,7 +116,7 @@ func Enroll(enFlags EnrollmentFlags) (*config.Config, error) {
 	if strings.TrimSpace(enFlags.KeyFile) != "" {
 		stat, err := os.Stat(enFlags.KeyFile)
```
plorenz: Are you going to add a check for the windows specific error?

dovholuknf: Was not planning on it - no. That's 'the other option' but there's no nice way to do it (imo) so checking stat for nil is just as good for me...

plorenz: ok, makes sense.

PR #23, opened 2020-01-28
get fix for ziti-foundation/GH-26

scareything:
Update ziti-foundation to get netfoundry/ziti-foundation#26

btw, is this is "right" way to propagate fixes through the dependency chain?


plorenz:
> Update ziti-foundation to get [netfoundry/ziti-foundation#26](https://github.com/netfoundry/ziti-foundation/issues/26)
> 
> btw, is this is "right" way to propagate fixes through the dependency chain?

Long term we'll want to push things up step by step. In the short term, while we don't have automated support, you only need to go step by step if you're making API changes, and even then, you only really need to do it in the projects that are affected by the API change. At least that's been my experience so far.

PR #29, opened 2020-02-10
Add versioning and tagging

plorenz:


ziti/sdkinfo/host.go
```diff
@@ -0,0 +1,48 @@
+/*
+	Copyright 2019 NetFoundry, Inc.
```
dovholuknf: time machine! :)

ziti/sdkinfo/host_unix.go
```diff
@@ -0,0 +1,32 @@
+// +build !windows
+
+/*
+	Copyright 2019 NetFoundry, Inc.
```
dovholuknf: 2020 - there's a few other spots to update from 2019

ziti/sdkinfo/host.go
plorenz: fixed

ziti/sdkinfo/host_unix.go
plorenz: fixed

PR #30, opened 2020-02-11
updates refs for yubikey fixes, error handling

andrewpmartinez:


ziti/enroll/enroll.go
```diff
@@ -269,31 +269,31 @@ func enrollOTT(token *config.EnrollmentClaims, cfg *config.Config, caPool *x509.
 		return err
 	}
 
-	respBody, err := ioutil.ReadAll(resp.Body)
```
andrewpmartinez: @dovholuknf I want you to look at this section. You were in here last and updated this logic - I want to make sure this is okay with what you saw when you were in here.

andrewpmartinez: Using gabs to parse the JSON allows us to not depend on ziti-edge.

dovholuknf: > @dovholuknf I want you to look at this section. You were in here last and updated this logic - I want to make sure this is okay with what you saw when you were in here.

we chatted. quick synopsis... @andrewpmartinez is gonna look into the causemessage/cause.message and add back the one that makes sense/is used.

ziti/enroll/enroll.go
```diff
@@ -269,31 +269,39 @@ func enrollOTT(token *config.EnrollmentClaims, cfg *config.Config, caPool *x509.
...
+	if err != nil {
+		return errors.Errorf("enroll error: %s: could not parse body: %s", resp.Status, body)
+	}
+
+	if jsonErr.Exists("error", "message") {
+		message := jsonErr.Search("error", "message").Data().(string)
+		code := jsonErr.Search("error", "code").Data().(string)
+
+		//todo: remove causeMessage support when removed from API
+		cause := "unspecified"
```
andrewpmartinez: Added cause handling back in here. I plan on removing `causeMessage` in netfoundry/ziti-edge#78

ziti/enroll/enroll.go
andrewpmartinez: Updated. See comment lower down.

PR #32, opened 2020-02-25
version 0.10

andrewpmartinez:
Talking w/ @plorenz he suggested for 0.10 that only `ziti-cmd`, `ziti-go-sdk` and `ziti-edge` need the `0.10` version marking and that `ziti-fabric` and `ziti-foundation` can be versioned separately.


andrewpmartinez:
Talking w/ @plorenz he suggested for 0.10 that only `ziti-cmd`, `ziti-go-sdk` and `ziti-edge` need the `0.10` version marking and that `ziti-fabric` and `ziti-foundation` can be versioned separately.

michaelquigley:
Managing versions on `ziti-fabric` and `ziti-foundation` separately seems fine. Those are internal dependencies for the others. Even eventually when there are external folks with dependencies on those things, versioning them independently of our "distribution version" will be fine.

PR #34, opened 2020-02-26
SDK: end-to-end encryption

ekoby:


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
 	github.com/michaelquigley/pfxlog v0.0.0-20190813191113-2be43bd0dccc
 	github.com/mitchellh/mapstructure v1.1.2
+	github.com/netfoundry/secretstream v0.1.1
```
andrewpmartinez: Were there reviews for this?

ziti/internal/edge_impl/conn.go
```diff
@@ -298,4 +382,14 @@ func (conn *edgeConn) newChildConnection(event *edge.MsgEvent) {
...
 	}
+
+	clientKey := message.Headers[edge.PublicKeyHeader]
+	if clientKey != nil {
+		err = edgeCh.establishCrypto(conn.keyPair, clientKey, true)
+		if err != nil {
+			logger.Errorf("failed to establish crypto session %v", err)
+		}
+	} else {
+		logger.Warnf("client did not send its key. connection is not end-to-end encrypted")
```
andrewpmartinez: What is the scenario where this would be ok?

go.mod
ekoby: Yes, https://github.com/netfoundry/secretstream/pull/1

ziti/internal/edge_impl/conn.go
ekoby: when the service is not hosted (egress from a router)

Issue #22, opened 2020-01-28
Handle reconnect/retry where appropriate, including unavailable edge-router and controller


plorenz:

plorenz:
* Handle listen to multiple edge routers.
* Allow configuring max number of connections to edge routers
* Allow configuring max time with no edge routers available before exiting
* Allow establishing new session, if existing session goes away
* Allow establishing new API session, existing API session goes away

* Ensure that terminators are cleaned up when edge routers goes down and comes back
* Ensure that terminators are cleaned up when SDK app closes abruptly 

* If client doesn't have access to service, it should stop listening return error
* If client can't establish or re-establish API session, it should stop listening and return error

PR #40, opened 2020-04-21
Make SDK more resilient. Fixes GH-22

plorenz:


go.mod
```diff
@@ -2,10 +2,11 @@ module github.com/netfoundry/ziti-sdk-golang
 
 go 1.14
 
-//replace github.com/netfoundry/ziti-foundation => ../ziti-foundation
+// replace github.com/netfoundry/ziti-foundation => ../ziti-foundation
 
 require (
 	github.com/Jeffail/gabs v1.4.0
+	github.com/cenkalti/backoff/v4 v4.0.2
```
ekoby, line 9: I didn't find where this was used

ziti/ziti.go
```diff
@@ -553,18 +639,261 @@ func (context *contextImpl) Close() {
...
+	}
+}
+
+func (mgr *listenerManager) createSessionWithBackoff() {
+	expBackoff := backoff.NewExponentialBackOff()
+	expBackoff.InitialInterval = 50 * time.Millisecond
+	expBackoff.MaxInterval = 10 * time.Second
+	expBackoff.MaxElapsedTime = mgr.options.GetConnectTimeout()
+
+	_ = backoff.Retry(mgr.createSession, expBackoff)
```
plorenz, line 839: new backoff library is used here

go.mod
plorenz: I put comments on the two places it is used

ekoby: thanks, ziti.go was collapsed in the view so my search returned nothing

PR #43, opened 2020-04-22
Fix close

plorenz:
Depends on https://github.com/netfoundry/ziti-foundation/pull/60/files

Fixes issue where close from remote wouldn't fully close local (reads would never return) and local closes would potentially cause race conditions, by closing sequencer which is only supposed be closed by the producer

### DIFF
--- a/ziti/internal/edge_impl/conn.go
+++ b/ziti/internal/edge_impl/conn.go
@@ -264,6 +264,7 @@ func (conn *edgeConn) Read(p []byte) (int, error) {
 			conn.closed.Set(true)
 			return 0, io.EOF
 		} else if err != nil {
+			log.Debugf("unexepcted sequencer err (%v)", err)
 			return 0, err
 		}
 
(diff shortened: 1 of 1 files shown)


ziti/internal/edge_impl/conn.go
```diff
@@ -311,7 +316,21 @@ func (conn *edgeConn) Read(p []byte) (int, error) {
...
-	return conn.close(false)
+	event := &closeConnEvent{
+		conn:        conn,
+		remoteClose: false,
+		errorC:      make(chan error, 1),
+	}
+	conn.msgMux.Event(event)
+	select {
+	case err, ok := <-event.errorC:
+		if ok && err != nil {
```
dovholuknf, line 327: is it possible to not be ok and have an err?

plorenz: It's redundant. I cleaned it up

PR #44, opened 2020-04-30
Add precedence to listen options

plorenz:


ziti/edge/messages.go
```diff
@@ -42,11 +42,18 @@ const (
 	SessionTokenHeader = 1002
 	PublicKeyHeader    = 1003
 	CostHeader         = 1004
+	PrecedenceHeader   = 1005
+
+	PrecedenceDefault  Precedence = 0
```
andrewpmartinez, line 47: 255 values of precedence is enough?

plorenz: Don't expect more than 3 :)

PR #46, opened 2020-05-06
Enable setting precedence/cost for graceful shutdown. Log message lev…

plorenz:
…el cleanup

### DIFF
--- a/ziti/internal/edge_impl/conn.go
+++ b/ziti/internal/edge_impl/conn.go
@@ -73,7 +73,7 @@ func (conn *edgeConn) Write(data []byte) (int, error) {
 func (conn *edgeConn) Accept(event *edge.MsgEvent) {
 	conn.TraceMsg("Accept", event.Msg)
 	if event.Msg.ContentType == edge.ContentTypeDial {
-		pfxlog.Logger().WithFields(edge.GetLoggerFields(event.Msg)).Info("received dial request")
+		pfxlog.Logger().WithFields(edge.GetLoggerFields(event.Msg)).Debug("received dial request")
 		go conn.newChildConnection(event)
 	} else if event.Msg.ContentType == edge.ContentTypeStateClosed && event.Seq == 0 {
 		_ = conn.close(true)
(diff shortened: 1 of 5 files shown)


ziti/internal/edge_impl/conn.go
```diff
@@ -214,18 +209,18 @@ func (conn *edgeConn) Listen(session *edge.Session, serviceName string, options
 	conn.TraceMsg("listen", bindRequest)
 	replyMsg, err := conn.SendAndWaitWithTimeout(bindRequest, 5*time.Second)
 	if err != nil {
-		logger.Error(err)
+		logger.WithError(err).Error("failed to bind")
```
dovholuknf, line 212: you have the serviceName in this function - think these errors should all have that service name in them just in case it's useful down the road?

ziti/internal/edge_impl/listener.go
```diff
@@ -118,11 +143,47 @@ func NewMultiListener(serviceName string) MultiListener {
...
 	baseListener
+	listeners    map[edge.Listener]struct{}
+	listenerLock sync.Mutex
+}
+
+func (listener *multiListener) UpdateCost(cost uint16) error {
+	listener.listenerLock.Lock()
+	defer listener.listenerLock.Unlock()
+	for child := range listener.listeners {
+		if err := child.UpdateCost(cost); err != nil {
```
dovholuknf, line 160: if this fails mid-way through will some costs be updated and others not? should you try to update all of the listeners anyway and not fail-fast?

ziti/ziti.go
```diff
@@ -724,10 +724,13 @@ func (mgr *listenerManager) handleRouterConnectResult(result *edgeRouterConnResu
 			mgr.routerConnections[routerConnection.GetRouterName()] = routerConnection
 			go mgr.createListener(routerConnection, mgr.session)
 		}
+	} else {
+		pfxlog.Logger().Debugf("ignoring connection to %v, already have max connections %v", result.routerUrl, len(mgr.routerConnections))
```
dovholuknf, line 728: seems like a warning to me vs debug? just judging by the message this 'sounds' like a bad thing and i'd want to see that in the log?

ziti/internal/edge_impl/conn.go
plorenz: It's already there... the logger is defined above as

```
	logger := pfxlog.Logger().
		WithField("connId", conn.Id()).
		WithField("service", serviceName).
		WithField("session", session.Token)
```

ziti/ziti.go
plorenz: It's not a bad thing. The way it works as follows:

1. There's a configured number of max connections when hosting a service
2. When starting to host a service it will create a bind session. The bind session returns all the edge routers available to host the service.
3. The SDK will attempt to connect to all available edge routers simultaneously. As connections are made we'll try to listen over them, but only up the max number of configured connections. If we have 5 available edge routers and 3 max connections, we'll connect to the first three that come back. The next two will hit this condition, but it's fine.

ziti/internal/edge_impl/listener.go
plorenz: I considered this, but decided to fast fail for expediency. I'll go back and take a look again to see how much more complicated this makes the code. Part of my reasoning for doing it this way is that this is likely happening as part of a graceful shutdown, and if it fails, we're still shutting down, just with slightly less grace. So it's more of a nicety. But on the other hand, it's always nice to maximize gracefulness.

plorenz: Updated to try and update all

PR #47, opened 2020-05-07
Fix end-to-end encryption setup race condition

plorenz:


ziti/internal/edge_impl/conn.go
```diff
@@ -408,20 +447,19 @@ func (conn *edgeConn) newChildConnection(event *edge.MsgEvent) {
 	}
 
 	if startMsg.ContentType == edge.ContentTypeStateConnected {
+		if txHeader != nil {
+			newConnLogger.Debug("sending crpto header")
```
ekoby, line 451: spelling

plorenz: Fixed

PR #57, opened 2020-06-04
Fix rebind after edge router connection close. Fixes GH-56

plorenz:


ziti/edge/msg_mux.go
```diff
@@ -128,7 +127,7 @@ func (mux *MsgMux) handleEvents() {
 func (mux *MsgMux) ExecuteClose() {
 	mux.closed.Set(true)
 	for _, val := range mux.chanMap {
```
plorenz, line 129: edgeConn.Close() was doing async close via the msg mux queue. This would block because the queue was already executing the msg mux close. The conn close would time out, the msg mux close would finish and the conn close event would get dropped and never happen. Changing the API allows the MsgSinks to know they are being closed from the msg in the processing loop goroutine and can do close code directly.

ekoby:
should some rebinding changes be in the tunneler?

plorenz:
> should some rebinding changes be in the tunneler?

I think this will handle it. The one thing we could add would be some kind of notification to the application

PR #60, opened 2020-06-10
Service refresh

ekoby:
implement SDK service and sessions refresh
add option for refresh interval and service changes callback
fix improper session caching


ziti/config/options.go
```diff
@@ -0,0 +1,26 @@
+package config
```
plorenz, line 1: Should ListenOptions move here as well?

ekoby: I think so
related to this I'd like to refactor `listenManager` into a separate file at some point

PR #63, opened 2020-06-12
Don't cache bind sessions

plorenz:


ziti/ziti.go
```diff
@@ -680,10 +687,10 @@ func (context *contextImpl) refreshSession(id string) (*edge.Session, error) {
 	if err != nil {
 		return nil, err
 	}
-	return context.toSession("refresh", resp, 200)
+	return context.toSession("refresh", resp, 200, cache)
 }
 
-func (context *contextImpl) toSession(op string, resp *http.Response, expectedStatus int) (*edge.Session, error) {
+func (context *contextImpl) toSession(op string, resp *http.Response, expectedStatus int, cache bool) (*edge.Session, error) {
```
ekoby, line 693: I don't think the parameter is needed since session has the 'type' field and cache check can be done here

plorenz: Fixed

PR #70, opened 2020-06-25
Add simple echo example

dovholuknf:
usage: 
`
sever: echo.exe server -i server.json -s echosvc
client: echo.exe client -i client.json -s echosvc
`

Here's the server/client example in action:
image


ziti/ziti.go
```diff
@@ -447,6 +447,9 @@ func (context *contextImpl) GetServiceId(name string) (string, bool, error) {
 }
 
 func (context *contextImpl) GetService(name string) (*edge.Service, bool) {
+	if err := context.initialize(); err != nil {
```
dovholuknf, line 450: if you call GetService without this change you'll be surprised when you get no results.

dovholuknf:
> I like it. I do think some high level comments wouldn't hurt. And I'm just a little queasy about the executable name of "echo.exe".

Oh that's actually feedback is didn't even occur to me. Any ideas that wouldn't make you queasy?  I wanted a single executable to keep it easy... I had it called ziti-echo at first but it looked stuttery...

scareything:
> Any ideas that wouldn't make you queasy?

Nothing great comes to mind. simple-echo doesn't feel terrible though.

PR #75, opened 2020-07-31
handle 0-terminated strings

ekoby:


ziti/sdkinfo/host_unix.go
```diff
@@ -20,12 +20,21 @@ package sdkinfo
...
 	"golang.org/x/sys/unix"
+	"strings"
 )
 
+var nullTerm = string([]byte{0})
+
+// get string out of \0-terminated bytes
+func toString(b []byte) string {
+	s := strings.SplitN(string(b), nullTerm, 2)
+	return s[0]
```
dovholuknf, line 31: if len(s) is > 0 - i think there should be a warning here or something?

Issue #80, opened 2020-08-24
Do not call os.Exit


andrewpmartinez:
The Go SDK has a few cases where it will cause the hosting process to exit.

- Remove them
- Deal with situations where API sessions can be reused

PR #81, opened 2020-08-24
add graceful api session handling, reconnect #80

andrewpmartinez:
- API Sessions are maintained as long as they remain valid
- handle situations where a Session becomes invalid
- removes fatal calls


ziti/ziti.go
```diff
@@ -132,44 +134,22 @@ func (context *contextImpl) ensureConfigPresent() error {
...
-
-	// get services
-	if services, err := context.getServices(); err != nil {
-		return err
-	} else {
-		context.processServiceUpdates(services)
-	}
-
+	context.ctrlClt, err = api.NewClient(context.zitiUrl, context.id.ClientTLSConfig())
 	return nil
```
ekoby, line 153: return `err` from the line above?

ziti/ziti.go
```diff
@@ -295,30 +293,52 @@ func (context *contextImpl) Authenticate() error {
...
+		session, err := context.GetSession(serviceId)
 		if err != nil {
 			return nil, err
 		}
-		conn, err = context.dialSession(serviceName, ns)
-		if err != nil && attempt == 0 {
+		pfxlog.Logger().Infof("connecting via session id [%s] token [%s]", session.Id, session.Token)
+		conn, err = context.dialSession(serviceName, session)
+		if err != nil {
 			if strings.Contains(err.Error(), "closed") {
```
ekoby, line 340: this match seems very broad

andrewpmartinez: I don't know what it is for or why it is there. IIRC I set a breakpoint on it and it was never triggered.

PR #95, opened 2020-10-13
Ignore wss URLs

plorenz:


ziti/edge/api/client.go
```diff
@@ -307,5 +306,10 @@ func decodeSession(resp *http.Response) (*edge.Session, error) {
 		pfxlog.Logger().WithError(err).Error("failed to decode session response")
 		return nil, err
 	}
+
+	for _, edgeRouter := range session.EdgeRouters {
+		delete(edgeRouter.Urls, "wss")
```
rentallect, line 311: Please also add "ws". I am currently prototyping that transport on behalf of pure JS tls connections.

plorenz: done

rentallect (approved):

PR #101, opened 2020-10-20
add ability for enroller to emit EC|RSA keys

rentallect:


ziti/enroll/enroll.go
```diff
@@ -53,6 +54,7 @@ type EnrollmentFlags struct {
 	JwtString     string
 	CertFile      string
 	KeyFile       string
+	KeyType       string
```
dovholuknf, line 57: is this a 'type'? maybe this should be 'algorithm' instead of 'type' like this? are there only two types? i only know of two types but i an not cryptogropher :)

dovholuknf: maybe we could keep EC or RSA as shortcuts to a 'default' algorithm so that all this can basically stay the same (to keep it simple/ship it, yet still fwd compatible?), just rename it to KeyType and any other input for now is 'errored' out?

dovholuknf: lol rename it to Algorithm or KeyAlgorithm or something i mean... or just something not 'KeyType'? :)

rentallect: I can rename to `KeyAlg` if that makes more sense.

rentallect: fwiw, Go docs actually refers to this stuff with the word `kind`, as in, "This kind of key ..."

dovholuknf (approved):

PR #104, opened 2020-10-22
Allow nil headers

plorenz:


rentallect (approved):

ziti/edge/conn.go
```diff
@@ -131,8 +131,10 @@ func (ec *MsgChannel) WriteTraced(data []byte, msgUUID []byte, hdrs map[int32][]
 		msg.Headers[UUIDHeader] = msgUUID
 	}
 
-	for k, v := range hdrs {
-		msg.Headers[k] = v
+	if hdrs != nil {
```
ekoby, line 134: is this needed? range over nil map is safe -- skips the loop

plorenz: Yeah, I could have sworn I hit a panic on this, but you are correct, it should not panic. If I see the panic again, I'll dig in deeper.

ziti/edge/impl/conn_test.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+	}()
+
+	req := require.New(b)
+
+	req.NoError(mux.AddMsgSink(conn))
+
+	data := make([]byte, 1024)
+
+	b.ResetTimer()
```
michaelquigley, line 203: Could be wrong, but I believe the _best_ way to write a golang benchmark is to engineer the actual `Benchmark` function, so that this stuff is done in an `init` or some other setup infrastructure, because even though this stuff is outside the `i < b.N` loop, it'll still affect the results a little bit.

plorenz: the docs here: https://golang.org/pkg/testing/ point to `b.ResetTimer()` as a way to avoid the issue.

PR #109, opened 2020-11-12
adds windows server for openziti/edge#406

andrewpmartinez:


ziti/edge/posture/os.go
```diff
@@ -28,12 +29,23 @@ type OsInfo struct {
...
-	osType := runtime.GOOS
+	osType := "unknown"
 	osVersion := "unknown"
 
 	semVerParser := regexp.MustCompile(`^((\d+)\.(\d+)\.(\d+))`)
 
-	version, _ := host.KernelVersion()
+	_, family, version, _ := host.PlatformInformation()
+
+	if runtime.GOOS == "windows" {
```
dovholuknf, line 39: to lower case just in case?

dovholuknf (approved):

ziti/edge/posture/os.go
andrewpmartinez: Don't this it is necessary, formally defined values.

PR #118, opened 2020-12-08
Add ability to manually start or fail accepted edge connections

plorenz:
Also support router provided connect ids, which simplifies dialing sdk hosted endpoints

### DIFF
--- a/ziti/edge/impl/conn.go
+++ b/ziti/edge/impl/conn.go
@@ -44,16 +44,17 @@ func init() {
 
 type edgeConn struct {
 	edge.MsgChannel
-	readQ          sequencer.Sequencer
-	leftover       []byte
-	msgMux         edge.MsgMux
-	hosting        sync.Map
-	closed         concurrenz.AtomicBoolean
-	readFIN        concurrenz.AtomicBoolean
-	sentFIN        concurrenz.AtomicBoolean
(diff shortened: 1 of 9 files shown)


ziti/edge/impl/conn.go
```diff
@@ -507,32 +514,105 @@ func (conn *edgeConn) newChildConnection(message *channel2.Message) {
...
+		}
+	} else if err := self.conn.SendPrioritizedWithTimeout(reply, channel2.Highest, time.Second*5); err != nil {
+		logger.Errorf("Failed to send reply to dial request: (%v)", err)
+		return err
+	}
+
+	if self.txHeader != nil {
+		newConnLogger.Debug("sending crypto header")
+		if _, err := self.edgeCh.MsgChannel.Write(self.txHeader); err != nil {
+			newConnLogger.Errorf("failed to write crypto header: %v", err)
```
andrewpmartinez, line 612: I don't know if you are trying to standardized error output but this one is `%v` and not `(%v)`

andrewpmartinez (approved):

ziti/edge/impl/conn.go
plorenz: Did some tidying in the file. Where possible I used WithError instead of Errorf

PR #127, opened 2021-01-18
add support for appId and appVersion

andrewpmartinez:


ziti/ziti.go
```diff
@@ -57,6 +57,7 @@ type Context interface {
 	GetServiceId(serviceName string) (string, bool, error)
 	GetServices() ([]edge.Service, error)
 	GetService(serviceName string) (*edge.Service, bool)
+	SetAppInfo(appId, appVersion string)
```
ekoby, line 60: I don't think these should be per-context values, the app would be the same for all ziti contexts

andrewpmartinez: Ok

andrewpmartinez: Fixed, let me know.

plorenz (approved):

PR #130, opened 2021-01-20
Fix panic when passing in nil config. Move config types into client

plorenz:


ekoby (approved):

plorenz:
@andrewpmartinez so you're not confused, I grabbed a change from your PR (fixing the return type on GetSdkInfo())

PR #133, opened 2021-01-29
reworks posture responses

andrewpmartinez:
- send via bulk updates
- only send on change

### DIFF
--- a/ziti/edge/api/client.go
+++ b/ziti/edge/api/client.go
@@ -61,6 +61,7 @@ type RestClient interface {
 	CreateSession(svcId string, kind edge.SessionType) (*edge.Session, error)
 	RefreshSession(id string) (*edge.Session, error)
 	SendPostureResponse(response PostureResponse) error
+	SendPostureResponseBulk(responses []*PostureResponse) error
 
 	AuthenticateMFA(code string) error
 	VerifyMfa(code string) error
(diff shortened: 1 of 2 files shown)


ziti/edge/api/client.go
```diff
@@ -167,11 +169,39 @@ func (c *ctrlClient) CreateSession(svcId string, kind edge.SessionType) (*edge.S
 	return decodeSession(resp)
 }
```
plorenz, line 171: Do we need to fallback to individual submissions if the endpoint isn't supported? Could remember if you get a 404 and call the individual submission api instead.

andrewpmartinez: Sure should and now does.

ziti/edge/posture/cache.go
```diff
@@ -18,231 +18,321 @@ package posture
...
+
+	keys := cache.watchedProcesses.Keys()
+	for _, processPath := range keys {
+		cache.currentData.Processes.Set(processPath, Process(processPath))
 	}
+}
 
-	cache.MacAddresses = MacAddresses()
-	cache.Os = Os()
+// requires both slices to be sorted, returns true if slices are exact equal sets
```
plorenz, line 245: Can use stringz.EqualSlices

plorenz (approved):

ziti/edge/posture/cache.go
andrewpmartinez: Done and removed unused function.

PR #147, opened 2021-04-02
Get tags for current user

plorenz:


ziti/edge/types.go
```diff
@@ -26,10 +26,11 @@ import (
 )
 
 type CurrentIdentity struct {
-	Id                       string `json:"id"`
-	Name                     string `json:"name"`
-	DefaultHostingPrecedence string `json:"defaultHostingPrecedence"`
-	DefaultHostingCost       uint16 `json:"defaultHostingCost"`
+	Id                       string                 `json:"id"`
+	Name                     string                 `json:"name"`
+	Tags                     map[string]interface{} `json:"tags"`
```
andrewpmartinez, line 31: As a warning, this isn't planned to survive the API split. Tags will only be on the management API and the SDK will operate on the Client API.

andrewpmartinez (changes requested):
See comment.

ziti/edge/types.go
plorenz: Admins can set source IP by tag, which is then used by tunnelers

andrewpmartinez: Is there a reason this data isn't added to the tunneler service that is meant to deal with IP management?

plorenz: I think that's a question for @scareything Not sure how ippool Icommunicates with sdk

andrewpmartinez: The issue with adding that data to tags remains the same, tags are meant for external 3rd party non-ziti components.

If we really want tunneler data in the model I suggest we add the fields properly to the model. I consider this an abuse of the intention of tags and sets us up for external non-ziti users of tags to accidentally misuse the field as there is no validation of tag data. Adding special validation to some tags but not others is not an option in my opinion.

I suggest one of the following in order of preference:

1) don't put the data here instead put it in our IP management capabilities
2) if more tunneler data will be added to the data model add a `tunneler` subfield that holds `sourceIP` and other future properties or prefix it with `tunneler` as in `tunnelerSourceIp`. Former over later.

andrewpmartinez: For help w/ the swagger this is the definition that would be needed:

```
tunnelerOptions:
  type: object
  required:
    - sourceIp
  properties:
    - sourceIp
      type: string
      format: ipv4


identityDetail:
  ...
  properties:
    ...
    - tunneler:
        $ref: '#/definitions/tunnelerOptions'

```

smilindave26: SDKs are meant to be used by external 3rd parties (we encourage others at NetFoundry to think of the tunnelers like 3rdp party developed apps).  As this use case demonstrates, we need a way to generically associate identity-specific extensions (e.g., tags) with services

andrewpmartinez: I can understand that viewpoint and for external code that we don't write control, I agree that they can use tags as we can't prescribe to them what they do/don't do with tags.  Tags were never meant to go to the client, they did so by accident. 

Consider the following:

1) having only one place for tags means that an implementor can't hide custom data from endpoints when there maybe be a good reason to
2) we should most likely introduce `client/identity/endpointTags`  or some other aptly named value that is sent only to clients, giving implementors the flexibility to selectively choose what values are exposed to endpoints.

Returning to the tunneler sourceIp value being in tags:

If the tunneler values are going to read and used by the tunneler SDK, the functionality of the field is not an external application, it is our tunneler SDK and part of internal development. If the sourceIp values is passed out of our SDK for external app use then I posit the following:

It is problematic to provide the tunnelers that are branded as  `Ziti` (and by extension the ZDE/ZME) using tags for values. That is because an implementor will question why the "Ziti" app broke because a set of values meant for external implementation broke it by setting the values incorrectly, removing them, etc.

I understand how we sell the tunneler as "3rd party" but it has implications for end-users who might not make that distinction because of the branding on the ZDE/ZME. We should consider and accept the future customer support issues this decision can cause upfront and if we are still fine with it move forward. I think we will look ridiculous in that situation and I'd rather avoid it through design rather than documentation.

smilindave26: The key requirement is "we need a way to generically associate identity-specific extensions (e.g., tags) with services" - for SDK users. We can separate the question of whether NetFoundry-developed code/extensions should take advantage of the feature.

smilindave26: (but hard-coding "tunneler" attributes in our data model feels wrong to me - like a violation of open/closed principle)

scareything:
ippool is another method of retrieving ip addresses, and does not rely on tags at all. The work is still in flight, but basically ippool is a ziti service that tunnelers dial to get ips. So the sourcIp field for a service that uses ippool might look like `$ipPool[serviceName]` (vs `$tunneler_id.tag[tagName]`)

ziti/edge/types.go
plorenz: My preference would be to add an AppData map[string]interface{} to Identity that we, and third party SDK users, could use for arbitrary data like this. We already have configs for services, so I don't think we need anything extra there. The only other entity exposed to the SDK is edge routers, but those aren't really surfaced past the sdk, so I don't think that would need it either.

michaelquigley: We seem to keep losing the plot with this. I do think we should be using our own extension mechanisms to support the non-core parts of our stack. If not "tags", then a generic "data" namespace (like Paul is describing) probably makes sense.

We probably should add it for routers though also. I'd rather see us do this more consistently rather than calling it one thing for services, another name for data on identities, and yet another name some day when we add it for routers.

michaelquigley: But also, using "tags" for arbitrary data is a fairly common extension approach for lots of things. There's really not much difference between "tags" and "appdata" other than intent, and when it comes to extension points, you can't really prescribe the intent anyway.

plorenz: I don't mind adding it for routers and giving it the same name and structure. I think it's ok having it separate from service configs because configs are a separate data structure which is associated to services and identities. I'd rather keep something simpler than configs for identity and edge routers b/c configs feel like overkill for the use case.

michaelquigley: Sounds reasonable to me. I just don't think it makes sense to layer in data into the core for every part of our stack, just because it's our code. And I also think if it makes sense to establish a pattern (like `AppData`) then we probably ought to add it for routers as well.

andrewpmartinez: I am for  `AppData`  or it by any other name. See my comment about adding `client/identity/endpointTags` above, that is exactly what I was attempting to suggest.

The number one thing I am trying to avoid in this PR is using what was intended as management API data for arbitrary endpoint data. I believe they should be separate for security reasons and because of the API split. It sounds like `tags` and when the API splits it literally can be named `tags` as there won't be anything like it in the client/endpoint API.

Secondarily I was questioning if we need to support this in the controller at all. It seems like a perfect fit for the IP management service that @scareything was working on. If it can't go there then it can go in the aforementioned map[string]interface{} as long as we are fine with people accidentally breaking it and then questioning why, what will appear to them as our internal data, is in a space meant for "their" data. If we are fine with that risk - that is all I need.

plorenz: I don't think everyone is going to want to run the ip management stack. It may come for free initially, but may also be split out and be an optional component to maintain. AppData should be documented not as a place for customer data, but as a place for application specific data, with the tunneler as an example and indicate that keys should be namespaced. That won't necessarily prevent all confusion, but hopefully most of it.

andrewpmartinez: I currently assume that `appData` will only be editable by admins and identities only have read access.

plorenz: @andrewpmartinez I think that's a good assumption

plorenz: Updated PR to use AppData. Work to support it in edge will be in a separate PR

andrewpmartinez (approved):
Nice!

PR #148, opened 2021-04-05
normalize file paths for posture checks 

dovholuknf:
This will allow case insensitive matching in windows and it will also allow for extra slashes to be fatfingered and still match

### DIFF
--- a/ziti/edge/posture/posture_windows_test.go
+++ b/ziti/edge/posture/posture_windows_test.go
@@ -1,0 +1,35 @@
+// +build windows
+
+package posture
+
+import (
+	"testing"
+)
+
+func TestRunningProcess(t *testing.T) {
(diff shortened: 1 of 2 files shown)


ziti/edge/posture/posture_windows_test.go
```diff
@@ -0,0 +1,50 @@
...
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+package posture
+
+import (
+	"testing"
+)
+
+func TestRunningProcess(t *testing.T) {
```
andrewpmartinez, line 24: You can nest tests to group them.

```
func TestWhatever(t *testing.T){
  t.Run("inner test 1", t *testing.T)
}
```

andrewpmartinez (approved):

plorenz (approved):

PR #151, opened 2021-04-19
Add error code header

plorenz:


andrewpmartinez (approved):

ekoby:
should we define possible error codes as well?

PR #167, opened 2021-07-02
Fix id check

plorenz:


ziti/edge/msg_mux.go
```diff
@@ -62,18 +62,20 @@ type CowMapMsgMux struct {
...
 	nextId := atomic.AddUint32(&mux.nextId, 1)
-	valid := false
 	sinks := mux.getSinks()
-	for !valid {
-		_, found := sinks[nextId]
-		valid = found || nextId <= mux.maxId
-
-		if !valid {
+	for {
+		if _, found := sinks[nextId]; found {
```
dovholuknf, line 67: it'd be helpful for a reviewer (like me) who doesn't remember what was going on here to understand 'why' this came about. did we end up rolling over maxuint32 / 2? If so - do we also need to make sure no actual "really old" id's are still hanging around and thus clashing? i don't remember the back story on this and when i looked for 'why' I didn't find it

plorenz: Yes, this is trying to handle both. If we go out of the range, reset to the beginning of the range. It also checks to ensure that any ID being used isn't already in use

dovholuknf (approved):

ziti/edge/msg_mux.go
plorenz: Previously there was a bug that allowed already in use ids because of bad boolean logic

PR #182, opened 2021-08-07
made err logging more consistent across zping-client/server and caugh…

r-caamano:
…t unhandled err in server.

### DIFF
--- a/example/zping-client/zping-client.go
+++ b/example/zping-client/zping-client.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"github.com/openziti/sdk-golang/ziti"
 	"github.com/openziti/sdk-golang/ziti/config"
+	"github.com/sirupsen/logrus"
 	"html"
 	"math"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


example/zping-client/zping-client.go
```diff
@@ -135,8 +136,8 @@ func main() {
 		//fmt.Println(file)
 		configFile, err := config.NewFromFile(file)
 		if err != nil {
```
plorenz, line 138: We usually use a pattern like: `logrus.WithError(err).Error("Problem loading config file")`

example/zping-server/zping-server.go
```diff
@@ -72,26 +73,29 @@ func main() {
 		file := *configPtr
 		configFile, err := config.NewFromFile(file)
 		if err != nil {
-			logger.Errorf("Problem loading configfile: %+v\n", err)
-			panic(err)
+			logrus.Errorf("Problem loading configfile: %+v\n", err)
```
plorenz, line 76: `os.Exit(1)`?

r-caamano: If you don't use panic the logrus just print and continues

example/zping-client/zping-client.go
r-caamano: This is pattern used in chat-server
if err != nil {
		logrus.Errorf("Error binding service %+v", err)
		panic(err)
	}

r-caamano: Will change to pattern as suggested

PR #187, opened 2021-08-11
document reflect example in docker/k8s

qrkourier:


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/openziti/sdk-golang
 go 1.16
 
 //replace github.com/openziti/foundation => ../foundation
```
plorenz, line 5: You can't replace a project within itself. Can we drop this comment?

qrkourier: @plorenz We can drop it. I left it in because I was troubleshooting building the project in a repo fork, something that I expect is typical for community contributors. I was able to build it with this `replace` directive, but did not wish to persist that change in the upstream. There's probably a better way.

plorenz: If I check out out your fork branch, it builds ok without that. I'm not clear on what problem it addressed, can you share some more details?

qrkourier:
I believe this branch got lost in the shuffle and is superseded by Geoff's more recent work in the same area.

PR #196, opened 2021-09-08
Add get service terminators. Add SourceIdentifier to conn. Cleanup ca…

plorenz:
…ll example

### DIFF
--- a/example/call/call.go
+++ b/example/call/call.go
@@ -18,11 +18,13 @@ package main
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/sdk-golang/ziti"
 	"github.com/openziti/sdk-golang/ziti/edge"
 	"github.com/sirupsen/logrus"
+	"io"
 	"os"
 	"strings"
 	"time"
@@ -204,6 +206,10 @@ func (app *callApp) consoleIO() {
 		fmt.Printf("%v > ", app.identity)
 		line, err := reader.ReadString('\n')
 		if err != nil {
+			if errors.Is(err, io.EOF) {
+				fmt.Println("\ngoodbye")
+				os.Exit(0)
+			}
 			panic(err)
 		}
 		line = strings.TrimSpace(line)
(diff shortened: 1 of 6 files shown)


ziti/edge/types.go
```diff
@@ -92,6 +92,13 @@ type Service struct {
 	Tags           map[string]string                 `json:"tags"`
 }
 
+type Terminator struct {
+	Id        string `json:"id"`
+	ServiceId string `json:"serviceId"`
+	RouterId  string `json:"routerId"`
+	Identity  string `json:"Identity"`
```
dovholuknf, line 99: why does this return an identity and not just the id?

dovholuknf: maybe it's just misnamed? also is the casing difference historical? (capital I)

plorenz: This is that whole terminator identity vs edge identity thing. It's the terminator identity, which we should rename to something less confusing. But for now, it's still Identity

ziti/edge/api/client.go
```diff
@@ -754,6 +755,77 @@ func (c *ctrlClient) GetServices() ([]*edge.Service, error) {
...
+	servReq, _ := http.NewRequest("GET", c.zitiUrl.ResolveReference(terminatorsUrl).String(), nil)
+
+	if c.apiSession.Token == "" {
+		return nil, errors.New("apiSession apiSession token is empty")
+	} else {
+		pfxlog.Logger().Debugf("using apiSession apiSession token %v", c.apiSession.Token)
+	}
+	servReq.Header.Set(constants.ZitiSession, c.apiSession.Token)
+	pgOffset := 0
+	pgLimit := 500
```
dovholuknf, line 772: nit: i couldn't find a const to reference in the sdk for this. should it be hardcoded like this? at least extract this to a const since it's in this file in 2 places now?

example/call/call.go
```diff
@@ -176,14 +195,28 @@ func (app *callApp) run() {
 }
 
 func (app *callApp) consoleIO() {
+	time.Sleep(250 * time.Millisecond)
```
dovholuknf, line 198: maybe a comment as to what that's for? sticks out when reviewing :)

dovholuknf (approved):
lgtm - maybe wait a day or two in case anyone else wants to weigh in?

ziti/edge/api/client.go
plorenz: Made paging explicit, so user can control it.

example/call/call.go
plorenz: Done

andrewpmartinez (approved):

PR #224, opened 2022-02-01
Update to use channel library

plorenz:


ziti/edge/impl/listener.go
```diff
@@ -114,7 +114,7 @@ func (listener *edgeListener) updateCostAndPrecedence(cost *uint16, precedence *
 	logger.Debug("sending update bind request to edge router")
 	request := edge.NewUpdateBindMsg(listener.edgeChan.Id(), listener.token, cost, precedence)
 	listener.edgeChan.TraceMsg("updateCostAndPrecedence", request)
-	return listener.edgeChan.SendWithTimeout(request, 5*time.Second)
+	return request.WithTimeout(5 * time.Second).SendAndWaitForWire(listener.edgeChan.Channel)
```
dovholuknf, line 117: all these 5 * time.second seem like they could be a constant or a function or 'anything else' maybe?

dovholuknf (approved):
i'd really like to see all those "5 * time.second" (in the whole pr) addressed but... lgtm

plorenz:
> i'd really like to see all those "5 * time.second" (in the whole pr) addressed but... lgtm

Yeah, I'd like to do a pass on figuring out how to do timeouts better. Make them more configurable and pass them through the server stack, so we can minimize work that's going to end up ignored.

PR #227, opened 2022-02-10
Prometheus Reflect Example

nf-npieros:
Updates to the reflect demo to add an example of adding custom prometheus metrics and exposing them over ziti


plorenz (changes requested):
Let's make the reflect example it's own go module so we don't clutter the sdk with extra dependencies. It can stay right where it is, it just needs its own go.mod

dovholuknf:
closing as stale.  we can reopen if needed

PR #240, opened 2022-04-07
Make edge router connection messages debugs

plorenz:


dovholuknf (approved):
i was promised one line

PR #243, opened 2022-04-13
fix potential leftover accounting problems

ekoby:
switch to use `bytes.Buffer` for read buffer instead of slice manipulation

### DIFF
--- a/ziti/edge/impl/conn.go
+++ b/ziti/edge/impl/conn.go
@@ -17,6 +17,7 @@
 package impl
 
 import (
+	"bytes"
 	"fmt"
 	"github.com/openziti/foundation/util/info"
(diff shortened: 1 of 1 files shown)


andrewpmartinez (approved):

plorenz (commented):
New code looks reasonable. It's simpler at the cost of an extra write. I can't see how the old code was broken. Can you explain how the old code was broken (or better yet add a test that demonstrates it)?

ekoby:
> New code looks reasonable. It's simpler at the cost of an extra write. I can't see how the old code was broken. Can you explain how the old code was broken (or better yet add a test that demonstrates it)?

I believe the old code was using `cap()` in places where `len()` should've been used

PR #244, opened 2022-04-19
Fix edge test breakage

plorenz:


ziti/edge/impl/conn.go
```diff
@@ -423,15 +419,18 @@ func (conn *edgeConn) Read(p []byte) (int, error) {
...
-						len(msg.Body), msg.Headers, err)
+					log.WithFields(edge.GetLoggerFields(msg)).Errorf("crypto failed on msg of size=%v, headers=%+v err=(%v)", len(msg.Body), msg.Headers, err)
 					return 0, err
 				}
 			}
-			conn.readBuffer.Write(d)
-			read, _ = conn.readBuffer.Read(p)
-			log.Debugf("reading %d bytes, %d bytes readBuffer", read, conn.readBuffer.Len())
-			return read, nil
+			if len(d) <= len(p) {
```
ekoby, line 426: this could be collapsed into a single `copy()` to simplify logic
```
n = copy(p, d)
conn.leftover = d[n:]
```

ekoby (approved):

ziti/edge/impl/conn.go
plorenz: Updated and tested

PR #247, opened 2022-04-22
allow specifying the identity to dial

dovholuknf:
quick PR to allow zcat sample to understand how to dial an identity


example/zcat/zcat.go
```diff
@@ -61,7 +61,7 @@ var root = &cobra.Command{
 			// let logrus do its own thing
 		}
 	},
```
plorenz, line 63: Should be `cobra.RangeArgs(1,2)`

plorenz (approved):

example/zcat/zcat.go
dovholuknf: sure. fixed

PR #254, opened 2022-05-18
Updatemod

r-caamano:
Updates to readme and go.mod for sdk updates


r-caamano:
Will resubmit with GPG signed commit

PR #284, opened 2022-08-17
update to run on ubuntu-latest and with go-1.19

plorenz:


.github/workflows/mattermost-ziti-webhook.yml
```diff
@@ -1,24 +1,31 @@
 name: mattermost-ziti-webhook
-on: 
-  create:
-  delete:
+on:
   issues:
   issue_comment:
   pull_request_review:
+    types: [submitted]
```
smilindave26, line 6: The `if` check (line 24) will prevent any `pull_request_review` other than `approved` from posting to Mattermost

smilindave26: nvm - I got that wrong... (the if check is for state, not action...)

andrewpmartinez (approved):

PR #332, opened 2022-11-04
SDK intercepts

ekoby:
make intercept configurations first class objects in SDK:
- add `ziti.Context.DialAddr(network, address) (net.Conn, error)`
- add `ziti.Dialer` that selects the best matching service from loaded identities and dials with appropriate `app_data`


ziti/contexts.go
```diff
@@ -0,0 +1,70 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+)
+
+var contexts = new(sync.Map)
+
+func init() {
+	ids := os.Getenv("ZITI_IDENTITIES")
```
dovholuknf, line 16: imo. sdk's should do 'magic' with environment variables like this.it should be explicitly initialized

ziti/contexts.go
```diff
@@ -0,0 +1,70 @@
...
+
+		_, err := LoadContext(id)
+		if err != nil {
+			pfxlog.Logger().Errorf("failed to load config from '%s'", id)
+			continue
+		}
+	}
+}
+
+func LoadContext(config_ string) (Context, error) {
```
dovholuknf, line 34: seems like this should be put next to `NewContextWithConfig` (ziti.go) and should also get a "WithOpts" example. Overall, I still don't think it belongs here, although I recognize it's convenient. it seems more like 'example code' to me.

ziti/ziti.go
```diff
@@ -569,6 +594,75 @@ func (context *contextImpl) DialWithOptions(serviceName string, options *DialOpt
 
 	return nil, errors.Wrapf(err, "unable to dial service '%s'", serviceName)
 }
+func (context *contextImpl) GetServiceForAddr(network, hostname string, port uint16) (*edge.Service, int, error) {
+	var service *edge.Service
+	score := math.MaxInt
+	context.intercepts.Range(func(key, value any) bool {
+		intercept := value.(*edge.InterceptV1Config)
+		fmt.Println(intercept)
```
dovholuknf, line 602: remove fmt.Println ?

ziti/edge/types.go
```diff
@@ -170,3 +181,188 @@ func ApiResponseDecode(data interface{}, resp io.Reader) (*ApiResponseMetadata,
...
+	SourceIp    *string
+	DialOptions *InteceptDialOptions `json:"dialOptions"`
+	Service     *Service
+}
+
+type DomainName string
+
+func (dn DomainName) Match(hostname string) int {
+	if len(dn) < 3 {
+		return -1
```
dovholuknf, line 208: why the len < 3 check? a comment would be useful here

ziti/edge/types.go
```diff
@@ -170,3 +181,188 @@ func ApiResponseDecode(data interface{}, resp io.Reader) (*ApiResponseMetadata,
...
+	}
+
+	if dn[0] == '*' {
```
dovholuknf, line 211: is DomainName trimmed? this could be an annoying bug if whitespace creeps in

ziti/edge/types_test.go
```diff
@@ -53,3 +54,55 @@ func TestNetworkSessionDecode(t *testing.T) {
...
+	}{
+		{params: args{network: "udp", hostname: "plain.host.ziti", port: 80}, expected: -1},
+		{params: args{network: "tcp", hostname: "plain.host.ziti", port: 80}, expected: 0},
+		{params: args{network: "tcp", hostname: "foo.domain.ziti", port: 80}, expected: int(3 << 16)},
+		{params: args{network: "tcp", hostname: "foo.host.notziti", port: 80}, expected: -1},
+		{params: args{network: "tcp", hostname: "bar.domain.ziti", port: 1024}, expected: int(3<<16 | 1000)},
+		{params: args{network: "tcp", hostname: "100.64.255.1", port: 80}, expected: 0},
+		{params: args{network: "tcp", hostname: "100.64.255.1", port: 1443}, expected: 1000},
+		{params: args{network: "tcp", hostname: "100.64.0.1", port: 80}, expected: (32 - 10) << 16},
+		{params: args{network: "tcp", hostname: "100.64.10.1", port: 1443}, expected: (32-10)<<16 | 1000},
```
dovholuknf, line 98: there are no negative tests. can/will you add some? using "tcp4"/"tcp6" here, nils etc.

ziti/edge/types.go
```diff
@@ -170,3 +181,188 @@ func ApiResponseDecode(data interface{}, resp io.Reader) (*ApiResponseMetadata,
...
+		domain := string([]byte(dn)[1:])
+		if strings.HasSuffix(hostname, domain) {
+			return len(hostname) - len(domain)
+		} else {
+			return -1
+		}
+	} else {
+		if hostname == string(dn) {
```
dovholuknf, line 219: why the if, why not return the comparison of hostname to string(dn)? if there's a reason, I'd love to see a comment explaining why

ziti/edge/types.go
```diff
@@ -170,3 +181,188 @@ func ApiResponseDecode(data interface{}, resp io.Reader) (*ApiResponseMetadata,
...
+		self.cidr = cidr
+		return nil
+	}
+
+	if ip := net.ParseIP(v); ip != nil {
+		self.ip = ip
+		return nil
+	}
+
+	if len(v) < 3 {
```
dovholuknf, line 277: is 3 the magic number of "{character}.{character}" is that what the three represents?  I expect that's what's happening? A comment would be helpful here explaining that magic number

ziti/edge/types.go
```diff
@@ -170,3 +181,188 @@ func ApiResponseDecode(data interface{}, resp io.Reader) (*ApiResponseMetadata,
...
+		return errors.New("invalid address")
+	}
+
+	if v[0] == '*' && v[1] != '.' {
```
dovholuknf, line 281: might be worthwhile to indicate in the doc that this function only does the most basic of testing and doesn't verify the domain is valid whatsoever.

ziti/edge/types.go
```diff
@@ -170,3 +181,188 @@ func ApiResponseDecode(data interface{}, resp io.Reader) (*ApiResponseMetadata,
...
+
+func (self *ZitiAddress) UnmarshalJSON(data []byte) error {
+	var v string
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+	return self.UnmarshalText([]byte(v))
+}
+
+func (intercept *InterceptV1Config) Match(network, hostname string, port uint16) int {
```
andrewpmartinez, line 297: These scoring methods need to be documented.

I don't understand why anyone would want scoring involved w/ dialing a network service.

ziti/ziti.go
```diff
@@ -569,6 +594,75 @@ func (context *contextImpl) DialWithOptions(serviceName string, options *DialOpt
...
+		sc := intercept.Match(network, hostname, port)
+		if sc != -1 {
+			if score > sc {
+				score = sc
+				service = intercept.Service
+			}
+
+			if sc == 0 {
+				return false
+			}
```
andrewpmartinez, line 612: So if there are two intercepts with the same score, we return the first one we happen across. Not sure how often that will happen, but is it deterministic?

ziti/edge/types.go
ekoby: scoring is needed when the same target address can match multiple services/intercepts. Most specific match is selected

scareything: regarding doc, we wrote this up back when we added scoring to the tSDK: https://github.com/openziti/ziti-tunnel-sdk-c/blob/main/docs/intercept-address-matching.md

ziti/dialer.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, err
+	}
+
+	network = normalizeProtocol(network)
+
+	var ztx Context
+	var service *edge.Service
+	best := math.MaxInt
+	ForAllContexts(func(ctx Context) bool {
+		srv, score, err := ctx.GetServiceForAddr(network, host, uint16(port))
```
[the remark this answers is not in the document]
ekoby: added tests

ziti/edge/types.go
ekoby: domain name comes from config that is validated on create

ziti/edge/types.go
ekoby: not sure what you mean by comparison. return is an value is an int -- -1 not matching, or the score.

ziti/ziti.go
ekoby: added determinism by using service name -- alphabetically first service wins

scareything (approved):

ziti/ziti.go
```diff
@@ -59,11 +61,16 @@ type Context interface {
 	GetCurrentIdentity() (*edge.CurrentIdentity, error)
 	Dial(serviceName string) (edge.Conn, error)
 	DialWithOptions(serviceName string, options *DialOptions) (edge.Conn, error)
+	// DialAddr finds the service for given address and creates a connection to it
+	DialAddr(network string, addr string) (edge.Conn, error)
```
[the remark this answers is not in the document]
ekoby: the main goal here is to simplify zitification of exitsing app/libraries.

basically existing code would go from `net.Dial('tcp','foo.bar:80')` to `ziti.NewZitiDialer().Dial('tcp','foo.bar:80')`

dialing with service/options is still available if an app needs them. this is to simplify initial conversion without introducing boilerplate.

PR #336, opened 2022-11-19
add Reload function to Context and implement

dovholuknf:
Right now the lower identity library has a Reload function on Identity but the golang sdk doesn't expose a method to instruct the identity to Reload. For a short-lived certificate from something like SPIFFE or vault the ability to reload an identity which had new certs provisioned is needed. @andrewpmartinez thinks exposing this function in this way will take care of this requirement


plorenz (approved):
I would bump the minor version of the sdk up, since this is an SDK addition.

dovholuknf:
Cool, what's the expected/proper way to bump the minor?

plorenz:
> Cool, what's the expected/proper way to bump the minor?

Edit the version file in the root of the project. The major.minor are managed manually via that file. The patch is bump automatically by the release build.

plorenz (approved):

dovholuknf:
> Edit the version file in the root of the project. The major.minor are managed manually via that file. The patch is bump automatically by the release build.

perfect. done. thanks

image

PR #342, opened 2022-12-06
fixes #331 always fetch CA bundle on enrollment

### LINKED ISSUE #331: CA bundle isn't always fetched during enrollment
see discourse for more details https://openziti.discourse.group/t/ziti-edge-enrol-identity-not-inserting-ca-bundle/814/30

It seems that if you have a PKI for the edge controller API, and a DIFFERENT PKI for the edge data plane, the enrollment of an identity will succeed but the logic that determines if the ca bundle needs to be pulled down will not trigger, and then when the identity tries to connect it will not be able to successfully connect to the routers.

code is at https://github.com/openziti/sdk-golang/blob/main/ziti/enroll/enroll.go#L232-L258

andrewpmartinez:


ekoby (approved):

qrkourier:
I'm concerned that introducing this change will cause newly-created identities to have the `id.ca` property defined with the well-known bundle from the controller and that most SDKs will subsequently _insist_ that server certs are verifiable with that bundle because it is present and fail when the server cert is from a third-party that's not in the bundle, like when an alt cert from Let's Encrypt is presented.

PR #347, opened 2023-01-18
allows JWTs to be sent for authentication and API requests

andrewpmartinez:


ziti/options.go
```diff
@@ -19,9 +20,11 @@ type Options struct {
 	RefreshInterval time.Duration
 	OnContextReady  func(ctx Context)
 	OnServiceUpdate serviceCB
+	HttpApiRequest  func(r *http.Request)
```
dovholuknf, line 23: an Options struct takes a function that does http requests? That seems -- odd to me? particularly without any sort of doc indicating why. Is this intentional? I don't see `HttpApiRequest` used in this PR, did I miss it?

ziti/ziti.go
```diff
@@ -79,13 +79,16 @@ type Context interface {
...
-	// Add a Ziti MFA handler, invoked during authentication
+	// AddZitiMfaHandler adds a Ziti MFA handler, invoked during authentication
 	AddZitiMfaHandler(handler func(query *edge.AuthQuery, resp func(code string) error) error)
 	EnrollZitiMfa() (*api.MfaEnrollment, error)
 	VerifyZitiMfa(code string) error
 	RemoveZitiMfa(code string) error
 
 	Reload() error
+
+	SetAuthToken(token string)
```
dovholuknf, line 90: In my opinion, we're at the point where we need to add doc along with any functionality we expect people to use/consume. Can this get doc added along with 'how to use it' as well, please?

dovholuknf (approved):
so far so good - other than adding doc/examples. i'll leave it up to you if that's in this pr or not

ziti/options.go
andrewpmartinez: Oh, meant to remove that. That was first pass stuff.

ziti/ziti.go
andrewpmartinez: Got you loud and clear. Will address on the morrow.

PR #377, opened 2023-04-18
adds authenticator layer to the SDK for the client and management api

andrewpmartinez:
- remove conf sub modules, renames config.New* to ziti.NewConfig* removes conflicts with app that have their own conf module
- adds authenciation mechanisms for client/management APIs
- uses new authentication mechanisms for all auth paths, patches old identity.ID configuration for backwards compat


example/influxdb-client-go/go.work
plorenz, line 1: go.work and go.work.sum shouldn't be committed

dovholuknf: i agree. there's like 14 of those files to rewind

ziti/identity.go
```diff
@@ -0,0 +1,105 @@
...
+
+// LazyIdentity will delay calling identity.LoadIdentity(config) till it is first accessed.
+type LazyIdentity struct {
+	identity.Identity
+	*identity.Config
+	loadOnce sync.Once
+}
+
+func (self *LazyIdentity) load() {
+	var err error
```
dovholuknf, line 37: the error and log should go inside the 'once'

ziti/identity.go
```diff
@@ -0,0 +1,105 @@
...
+	"crypto/x509"
+	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/identity"
+	"sync"
+)
+
+var _ identity.Identity = &LazyIdentity{}
+
+// LazyIdentity will delay calling identity.LoadIdentity(config) till it is first accessed.
+type LazyIdentity struct {
```
dovholuknf, line 30: if we want this, i think it belongs in the identity project and not here

ziti/contexts.go
```diff
@@ -107,41 +104,21 @@ func NewContextWithOpts(cfg *config.Config, options *Options) (Context, error) {
 		closeNotify:       make(chan struct{}),
 	}
 
-	ztUrl, err := url.Parse(cfg.ZtAPI)
-
-	if err != nil {
-		return nil, err
+	if cfg.ID.Cert != "" && cfg.ID.Key != "" {
```
dovholuknf, line 107: what if we started to add godoc to public functions and start here? this little bit of logic is really important to understand as a consumer

ziti/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"encoding/json"
+	"github.com/openziti/identity"
+	"github.com/pkg/errors"
+	"os"
+)
+
+type Config struct {
+	//ZtAPI should be in the form of https://<domain>[:<port>]/edge/client/v1
```
dovholuknf, line 27: another opportunity along with "zt-session" for a constant imo...  i don't think we have a good place for constants like this though

management/api-client.go
```diff
@@ -0,0 +1,109 @@
...
+
+func NewApiComponents(httpClient *http.Client, clientApiUrl *url.URL) (*rest_management_api_client.ZitiEdgeManagement, *openapiclient.Runtime) {
+	apiClientRunTime := openapiclient.NewWithClient(clientApiUrl.Host, clientApiUrl.Path, rest_management_api_client.DefaultSchemes, httpClient)
+
+	apiClient := rest_management_api_client.New(apiClientRunTime, nil)
+
+	return apiClient, apiClientRunTime
+}
+
+func NewHttpComponents() (*http.Client, *http.Transport, *tls.Config) {
```
dovholuknf, line 77: i feel like these `NewApiComponents` is growing the api needlessly and adds confusion. would you consider making this private?

management/api-client.go
```diff
@@ -0,0 +1,109 @@
...
+	"time"
+)
+
+type AuthenticatedApiClient struct {
+	*ApiClient
+	ZtSession     string
+	Authorization []string
+}
+
+func (a AuthenticatedApiClient) AuthenticateRequest(request runtime.ClientRequest, _ strfmt.Registry) error {
```
dovholuknf, line 23: this function is misnamed imo. It's more like "add authentiation headers" but also - does this need to be public?

management/api-client.go
```diff
@@ -0,0 +1,109 @@
...
+
+func NewAuthenticatedApiClient(apiSessionToken string, apiClient *ApiClient) *AuthenticatedApiClient {
+	authenticatedClient := &AuthenticatedApiClient{
+		ApiClient: apiClient,
+		ZtSession: apiSessionToken,
+	}
+
+	apiClient.Runtime.DefaultAuthentication = authenticatedClient
+
+	return authenticatedClient
```
dovholuknf, line 43: this is _our_ api client. we _know_ that it'll always need to be authenticated. adding the word "authenticated" is entirely redundant to me for that reason. I think "NewApiClient" is more succinct and jsut as descriptive

management/authenticators.go
```diff
@@ -0,0 +1,329 @@
+package ziti
```
dovholuknf, line 1: these functions will be some of the first ones people explore, i'd really appreciate godoc on them once we agree on the api

management/api-client.go
dovholuknf, line 1: there's already an api-client.go in sdk-golang/ziti/api-client.go... why not move this stuff there?

management/authenticators.go
```diff
@@ -0,0 +1,329 @@
...
+type ApiAuthenticator interface {
+	Authenticate(clientApiUrl *url.URL) (*AuthenticatedApiClient, *rest_model.CurrentAPISessionDetail, error)
+	AuthenticateWithClient(apiClient *ApiClient) (*rest_model.CurrentAPISessionDetail, error)
+	NewApiClient(clientApiUrl *url.URL) *ApiClient
+	SetInfo(envInfo *rest_model.EnvInfo, sdkInfo *rest_model.SdkInfo)
+	SetConfigTypes(configTypes []string)
+	GetBase() *BaseAuthenticator
+}
+
+type IdentityProvider interface {
```
dovholuknf, line 28: same redefinition here. i expect this is just an oversight? i don't want to make that same comment but there are __a lot__ of functions/interfaces/etc that seem duplicative?

dovholuknf (changes requested):
I'm going to pause my review and ask for changes i've already commented on. Maybe we can talk in person? I don't think we should merge this as is.

example/influxdb-client-go/go.work
andrewpmartinez: I tried not having them, but I had issues w/ making the examples that have a go.mod file reference the version of the SDK they are included with. Without the go.work/sum files they reference older versions of the go SDK. Meaning when you are working on it, the examples don't break immediately.

plorenz: I think if that's the goal, we should put `replace` stanzas in the go.mod files, instead of using go.work. Thoughts?

andrewpmartinez: I'm not against `replace` statements. I thought they were deprecated.

management/api-client.go
andrewpmartinez: It is an interface implementation. I do not get the choice.

management/api-client.go
andrewpmartinez: Different types. One is `ziti.ApiClient` the other is `management.ApiClient` and they have different internal concretes due to the way go-swagger generates clients. I could use generics, but the cost-benefit is low.

management/api-client.go
andrewpmartinez: Not all API requests are authenticated and thus not all clients need to be authenticated.

example/influxdb-client-go/go.work
plorenz: They're not deprecated (as far as I can tell). I think they have different use cases. go.work is for managing a developer's local environment. `replace` is for everything else (like managing intra-project references inside a monorepo).

ziti/config.go
andrewpmartinez: The edge does have those constants already. 

The edge depends on the SDK, so they would have to be here or in some other repo that isn't the edge. It is also kinda weird to move them out of edge as that is where their primary use is.

ziti/identity.go
andrewpmartinez: Sure. This function used to return the error, so it was defined outside the Once. Now that it isn't doing that, it can move.

example/influxdb-client-go/go.work
andrewpmartinez: Will do that then.

ziti/identity.go
andrewpmartinez: Sure.

ziti/contexts.go
andrewpmartinez: Sure.

management/api-client.go
andrewpmartinez: mmmmm I can do something to reduce the exposed names. A sub-package would do the same thing.

management/authenticators.go
```diff
@@ -0,0 +1,329 @@
...
+	"github.com/go-openapi/strfmt"
+	"github.com/openziti/edge-api/rest_management_api_client/authentication"
+	"github.com/openziti/edge-api/rest_model"
+	"github.com/openziti/identity"
+	"github.com/openziti/sdk-golang/ziti"
+	"github.com/openziti/sdk-golang/ziti/sdkinfo"
+	"net/url"
+)
+
+type ApiAuthenticator interface {
```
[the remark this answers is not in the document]
andrewpmartinez: It is similar but the root issue is to go-swagger generated code. It isn't implemented in a way where I can reference both clients via an interface without defining an interface through wrapper structs. It is possible, it is just kinda ugly and that interface is largely useless because the generated clients have vastly different capabilities.

So there is a trade-off, dupe code with similar naming or wrapper structs with interfaces and generics.

management/authenticators.go
andrewpmartinez: De-duping this one is possible.

management/authenticators.go
andrewpmartinez: Sure sure.

dovholuknf (approved):

PR #382, opened 2023-04-25
Add support for secondary auth in IdentityCredentials

potto007:
~~WIP: exploring how to add Secondary Credentials support in the least obtrusive manner~~

### DIFF
--- a/edge-apis/credentials.go
+++ b/edge-apis/credentials.go
@@ -239,7 +239,7 @@ func NewDualAuthCredentials(config identity.Config, jwt string) *DualAuthCredent
 }
 
 func (c *DualAuthCredentials) Method() string {
-	return "dual-auth"
+	return "cert"
 }
 
 func (c *DualAuthCredentials) GetCaPool() *x509.CertPool {
--- a/edge-apis/credentials.go
+++ b/edge-apis/credentials.go
@@ -157,7 +157,9 @@ var _ Credentials = &IdentityCredentials{}
 
 type IdentityCredentials struct {
 	BaseCredentials
-	Identity identity.Identity
+	Identity   identity.Identity
+	JWT        string
+	JWTenabled bool
 }
 
 // NewIdentityCredentials creates a Credentials instance based upon and Identity.
@@ -165,6 +167,8 @@ func NewIdentityCredentials(identity identity.Identity) *IdentityCredentials {
 	return &IdentityCredentials{
 		BaseCredentials: BaseCredentials{},
 		Identity:        identity,
+		JWT:             "",
+		JWTenabled:      false,
 	}
 }
 
@@ -173,7 +177,14 @@ func NewIdentityCredentialsFromConfig(config identity.Config) *IdentityCredentia
 	return &IdentityCredentials{
 		BaseCredentials: BaseCredentials{},
 		Identity:        &identity.LazyIdentity{Config: &config},
+		JWT:             "",
+		JWTenabled:      false,
 	}
+}
+
+func (c *IdentityCredentials) AddJwt(jwt string) {
+	c.JWT = jwt
+	c.JWTenabled = true
 }
 
 func (c *IdentityCredentials) GetIdentity() identity.Identity {
@@ -197,6 +208,14 @@ func (c *IdentityCredentials) TlsCerts() []tls.Certificate {
 	return nil
 }
 
+func (c *IdentityCredentials) AuthenticateRequest(request runtime.ClientRequest, _ strfmt.Registry) error {
+	if c.JWTenabled {
+		return request.SetHeaderParam("Authorization", "Bearer "+c.JWT)
+	} else {
+		return nil
+	}
+}
+
 var _ Credentials = &JwtCredentials{}
 
 type JwtCredentials struct {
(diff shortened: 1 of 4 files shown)


andrewpmartinez:
Can you elaborate on your use case? Are you trying to figure out how to authenticate w/ certificates and also provide a JWT? (I mention this because this is the only case I can figure that would need a primary auth w/ and a secondary).

potto007:
> Can you elaborate on your use case? Are you trying to figure out how to authenticate w/ certificates and also provide a JWT? (I mention this because this is the only case that would need a primary auth w/ a secondary).

Hey @andrewpmartinez - yes, I am finishing up making authentication with primary and secondary work when secondary is an external jwt. I tested it against the Controller several weeks ago using Postman, and verified that the auth works when using a cert identity along with a JWT bearer on every request.

I realized I went a bit too far on this new type, and I'm pairing it down to just augment IdentityCredentials to include a JWT field.

potto007:
Cleaning up, removing the custom type and instead overloading the IdentityCredentials type with optional JWT header support.

potto007:
This update to sdk-golang now simply adds to the IdentityCredentials type. An example of its use can be seen in the zssh PR I made: https://github.com/openziti-test-kitchen/zssh/pull/31 

The existing functionality in Edge is such that, at the end of the `cert` auth method is processing, the code checks for an auth policy, and if the auth policy has external JWT as a secondary auth, it goes on to check for a token in the header, validates it, and succeeds if everything is in order.

andrewpmartinez (changes requested):
I think this would be better as an improvement to `BaseCredentials` instead and de-JWT'ed. Rather add a `Headers map[string][]string{}`. Then augment the BaseCredentials.AuthenticateRequest to blindly apply those headers to the `runtime.ClientRequest`. Then alter all the other credentials first to call `self.BaseCredentials.AuthenticateRequest(cr, r)`.

The result is that credentials instantiations can do something like:


```
jwtStr := getJwt()
cred := NewIdentityCrednetialFromConfig(config)
cred.Headers["authorization"][]string{"Bearter " + jwtStr}
```

...or any other header that comes along. 

Reasoning:

1) Placing the functionality here limits the applicability to other cred mechanisms
2) Other authentication mechanisms are coming that are not JWTs but have header implications

ziti/contexts.go
```diff
@@ -127,7 +128,16 @@ func NewContextWithOpts(cfg *Config, options *Options) (Context, error) {
 	}
 
 	if cfg.ID.Cert != "" && cfg.ID.Key != "" {
-		cfg.Credentials = edge_apis.NewIdentityCredentialsFromConfig(cfg.ID)
+		idCreds := edge_apis.NewIdentityCredentialsFromConfig(cfg.ID)
```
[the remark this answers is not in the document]
potto007: Gotcha - I had wondered why the implicit nature here - that background makes perfect sense.

potto007:
That would be so much nicer! I'll give that a shot.

potto007:
@andrewpmartinez I've refactored credentials.go based on your review.

andrewpmartinez:
> 

Gonna look at this today.

edge-apis/credentials.go
```diff
@@ -109,15 +124,43 @@ func (c *BaseCredentials) GetCaPool() *x509.CertPool {
 	return c.CaPool
 }
 
+// Method provides a base implementation to return the authentication method of a Credentials instance.
+func (c *BaseCredentials) Method() string {
```
andrewpmartinez, line 128: This method was purposefully not implemented as it stops its usage as a `Credentials` interface implementor. The BaseCredentials struct is meant to be embedded as a component of another struct that provides that method and fulfills the interface.

ziti/contexts.go
```diff
@@ -87,7 +87,12 @@ func NewContextWithOpts(cfg *Config, options *Options) (Context, error) {
 	}
 
 	if cfg.ID.Cert != "" && cfg.ID.Key != "" {
+		baseCreds := &edge_apis.BaseCredentials{}
```
andrewpmartinez, line 90: This if-block should not change. It is for backward compatibility for systems that rely only on the previous file configuration logic.

If an implementor wants to use the new credentials interface, I would expect them to create their own credentials object and include any headers on that field. I don't want to mix the usage of the `cfg.ID` field and the `cfg.Credentials`.

andrewpmartinez (changes requested):
See comments.

edge-apis/credentials.go
```diff
@@ -71,34 +78,42 @@ func getClientAuthInfoOp(credentials Credentials, client *http.Client) func(*run
...
+	// SdkInfo is provided during authentication to set SDK information about the client.
 	SdkInfo *rest_model.SdkInfo
 
-	//CaPool will override the client's default certificate pool if set to a non-nil value.
+	// CaPool will override the client's default certificate pool if set to a non-nil value.
 	CaPool *x509.CertPool
 }
 
+// NewCredentials creates a new BaseCredentials instance.
+func NewCredentials() *BaseCredentials {
```
andrewpmartinez, line 98: This should not exist.

example/zping/go.work.sum
```diff
@@ -33,6 +33,7 @@ github.com/cyberdelia/templates v0.0.0-20141128023046-ca7fffd4298c/go.mod h1:GyV
 github.com/deepmap/oapi-codegen v1.3.13/go.mod h1:WAmG5dWY8/PYHt4vKxlt90NsbHMAOCiteYKZMiIRfOo=
```
andrewpmartinez, line 33: This file should not be included in the pr.

ziti/contexts.go
potto007: I modified this conditional in an attempt to avoid side-effects. The problem is that BaseCredentials gets overwritten by the `NewIdentityCredentialsFromConfig` call.

edge-apis/credentials.go
potto007: This is how I was using it: https://github.com/openziti-test-kitchen/zssh/blob/37c5561e1747ea312584487ca6ede3502702713c/zsshlib/ssh.go#L329

The problem is that, short of that, one needs to create another Credentials type that doesn't match the intent, making it less expressive... for instance, I could change the accessor to perform the following:

```
conf := getConfig(f.ZConfig)
conf.Credentials = edge_apis.NewJwtCredentials("")
conf.Credentials.AddHeader("Authorization", "Bearer "+token)
```
which would work, but gives the misconception that the JwtCredential type is actually being leveraged or is necessary.

edge-apis/credentials.go
potto007: I've removed it. It doesn't need to be an interface so long as the SDK user knows to update credentials after the `NewContext` call.

ziti/contexts.go
potto007: I've reverted this to the original conditional. The issue this conditional presents is that it creates a side-effect on the Credentials. My update to the code was an attempt to allow for the existing workaround behavior without it causing strange behavior. 

The workaround is to ensure that the SDK docs make it explicit that the credentials are not stable until _after_ the NewContext call is made.

potto007: An example of what I'm talking about is here: https://github.com/openziti-test-kitchen/zssh/blob/d6425563e497e5f93b2225816cc4a2a753655aab/zsshlib/ssh.go#L329 -- I needed to move the update of Credentials to be _after_ the call to `NewContext` in order to avoid `NewContextWithOpts` from overwriting `BaseCredentials.Headers`

potto007:
@andrewpmartinez any thoughts?

andrewpmartinez (approved):

edge-apis/credentials.go
```diff
@@ -8,26 +8,33 @@ import (
...
-	//Method return the authentication necessary to complete an authentication request.
+	// Method returns the authentication necessary to complete an authentication request.
 	Method() string
 
-	//ClientAuthInfoWriter is used to pass a Credentials instance to the openapi runtime to authenticate outgoing
+	// AddHeader adds a header to the request.
+	AddHeader(key, value string)
+
+	// AuthenticateRequest authenticates an outgoing request.
+	AuthenticateRequest(runtime.ClientRequest, strfmt.Registry) error
```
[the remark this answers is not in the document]
potto007: Good point - suggest I rename?

PR #383, opened 2023-04-28
fixes updb enrollment, underlying error types changed

andrewpmartinez:
Instead of checking for specific error types, try blindly to ensure success.

### DIFF
--- a/ziti/enroll/enroll.go
+++ b/ziti/enroll/enroll.go
@@ -147,9 +147,9 @@ func EnrollUpdb(enFlags EnrollmentFlags) error {
 			allowedCerts = append(allowedCerts, xcert)
 			caPool.AddCert(xcert)
 		}
-		//try again
+
 		if err := enrollUpdb(enFlags.Username, enFlags.Password, enFlags.Token, caPool); err != nil {
-			return fmt.Errorf("unabled to enroll after fetching server certs: %v", err)
+			return fmt.Errorf("unable to enroll after fetching server certs: %v", err)
 		} else {
 			return nil
 		}
--- a/ziti/enroll/enroll.go
+++ b/ziti/enroll/enroll.go
@@ -139,26 +139,20 @@ func EnrollUpdb(enFlags EnrollmentFlags) error {
 	ztApiRoot := enFlags.Token.Issuer
 
 	if err := enrollUpdb(enFlags.Username, enFlags.Password, enFlags.Token, caPool); err != nil {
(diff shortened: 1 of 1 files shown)


ziti/enroll/enroll.go
```diff
@@ -139,26 +139,20 @@ func EnrollUpdb(enFlags EnrollmentFlags) error {
...
+		rootCaPool.AddCert(enFlags.Token.SignatureCert)
 
-		return fmt.Errorf("unhandled error: %v", err)
+		for _, xcert := range FetchCertificates(ztApiRoot, rootCaPool) {
+			allowedCerts = append(allowedCerts, xcert)
+			caPool.AddCert(xcert)
+		}
+		//try again
+		if err := enrollUpdb(enFlags.Username, enFlags.Password, enFlags.Token, caPool); err != nil {
+			return fmt.Errorf("unabled to enroll after fetching server certs: %v", err)
```
dovholuknf, line 152: i hate being 'unabled'! :) (typo)

ziti/enroll/enroll.go
```diff
@@ -139,26 +139,20 @@ func EnrollUpdb(enFlags EnrollmentFlags) error {
...
+		pfxlog.Logger().Debug("fetching certificates from server")
+		rootCaPool := x509.NewCertPool()
+		rootCaPool.AddCert(enFlags.Token.SignatureCert)
 
-		return fmt.Errorf("unhandled error: %v", err)
+		for _, xcert := range FetchCertificates(ztApiRoot, rootCaPool) {
+			allowedCerts = append(allowedCerts, xcert)
+			caPool.AddCert(xcert)
+		}
+		//try again
```
dovholuknf, line 150: this is not "trying again" now, it's trying for the first time

ziti/enroll/enroll.go
dovholuknf, line 1: Do we have any doc that states explicitly you will now need to include the entire chain in the cas store if you're using a third party 'verifiable' CA?  Before, it'd use the OS trust store and now it exclusively fetches the certificates every time, which is fine, but we need to make sure people know that they need to add the full chain of trust for this to work now, assuming I'm understanding this change correcctly

andrewpmartinez: This only changes the error detection. Prior to this, the code checked for specific error types, which have changed with golang upgrades. This change removes the type assertions and looks for any error.

Other than that, I do not believe any behavior has changed.

dovholuknf (approved):

PR #387, opened 2023-05-03
Formalizes SDK collection functionality into a reusable component.

andrewpmartinez:
- converts the current implementation to the DefaultCollection global and deprecates it.
- adds doc to missing areas


ziti/collection.go
```diff
@@ -0,0 +1,143 @@
...
+	cmap "github.com/orcaman/concurrent-map/v2"
+	"os"
+	"strings"
+)
+
+// An SdkCollection allows Context instances to be instantiated and maintained as a group. Useful in scenarios
+// where multiple Context instances are managed together. Instead of using ziti.NewContext() like functions, use
+// the function provided on this type to automatically have contexts added as they are created. If ConfigTypes
+// is set, they will be automatically added to any instantiated Context through `New*` functions.
+type SdkCollection struct {
```
dovholuknf, line 30: SdkCollection? or ContextCollection? Why "SDK" vs "Context"?

ziti/collection.go
```diff
@@ -0,0 +1,143 @@
...
+
+// RemoveById removes a context by its string id.  It is not closed or altered in any way.
+func (set *SdkCollection) RemoveById(id string) {
+	set.contexts.Remove(id)
+}
+
+// ForAll call the provided function `f` on each Context.
+func (set *SdkCollection) ForAll(f func(ctx Context) bool) {
+	set.contexts.IterCb(func(key string, ctx Context) {
+		_ = f(ctx)
```
dovholuknf, line 102: why return the bool at all only to ignore it here? i sorta feel like a boolean should be supplied indicating to process them all regardless of outcome, or exit early on f() failing ?

ziti/collection.go
```diff
@@ -0,0 +1,143 @@
...
+	})
+}
+
+// NewContextFromFile is the same as ziti.NewContextFromFile but will also add the resulting
+// context to the current collection.
+func (set *SdkCollection) NewContextFromFile(file string) (Context, error) {
```
dovholuknf, line 108: should thse be changed to just be a reader? that way i couldd new a context from an env var as easily as a file by making a file reader/string reader/whatever reader

ziti/config.go
```diff
@@ -43,13 +43,27 @@ type Config struct {
 	Credentials apis.Credentials `json:"-"`
 }
 
+// NewConfig will create a new Config object from a provided Ziti Edge Client API URL and identity configuration.
+// The Ziti Edge Client API is usually in the format of `https://host:por/edge/client/v1`.
```
dovholuknf, line 47: typo. `:por`--> `:port`

ziti/contexts.go
```diff
@@ -4,103 +4,39 @@
...
-	"sync"
+	"strconv"
 )
 
-var contexts = new(sync.Map)
+var idCount = 0
 
-func init() {
-	ids := os.Getenv("ZITI_IDENTITIES")
+func NewId() string {
```
dovholuknf, line 19: Public function without doc. It's pretty straightforward imo but maybe continue the good vibes of adding the godoc?

andrewpmartinez: added

ziti/config.go
andrewpmartinez: fixed.

ziti/collection.go
andrewpmartinez: Backward compatibility originally.

It is a difference in how the map implementations work for interaction. Here I'm using cmap, which allows us to use generics for type safety. 

Another difference is that the returning bool on the visiting function doesn't exist. The original type (`sync.Map`) used that value to terminate looping early, but it doesn't signify an error, only "stop iterating".

Thinking about this now though, I can not use the bool callback here and instead cover that use case in ForAllContexts() for backward compatibility. Simplifying the new API.

ziti/collection.go
andrewpmartinez: Length, but I can make it CtxCollection.

ziti/collection.go
andrewpmartinez: I added that info the the `CtxCollection` type declaration.

ziti/default_collection.go
```diff
@@ -0,0 +1,80 @@
...
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package ziti
+
+// Deprecated: DefaultCollection is deprecated and is included for legacy support.
```
dovholuknf, line 19: as part of this deprecation, maybe it makes sense to change
```
func NewDialerWithFallback(ctx context.Context, fallback Dialer) Dialer
```
and
```
func NewDialer() Dialer {
	return &dialer{}
}
```

and add the ziti context/CtxCollection into that private `dialer` that's returned. That way when `Dial(network, address string) (net.Conn, error)` is called, the ziti context is available to be used and `ForAllContexts` isn't necessary? That seems like it'd work nicely?

dovholuknf: Right now it's implicit, so when the consumer makes a new dialer, they aren't the one keeping/tracking the state. That'd be a downside, it is kinda convenient from the consuming side to not have to deal with it... dunno

dovholuknf (approved):
seems like it'd work out. let's find out! :)

PR #390, opened 2023-05-05
Add back filtering out of ws and wss urls that was lost in refactor

plorenz:


ziti/client.go
```diff
@@ -161,6 +161,11 @@ func (self *CtrlClient) GetSession(id string) (*rest_model.SessionDetail, error)
 		return nil, err
 	}
 
+	for _, er := range resp.Payload.Data.EdgeRouters {
```
andrewpmartinez, line 164: A comment on why in the code would be nice, but lgtm.

andrewpmartinez (approved):

ziti/client.go
plorenz: comment added :+1:

PR #403, opened 2023-05-21
Add support for dialing edge routers over websocket

smilindave26:
* Dials ws and tls by default 
* Dials only ws when built for WASM
* Updated edge-api to v0.25.21


plorenz (commented):
Do we want regular go apps dialing ws endpoints? Should we make protocol selection configurable?

smilindave26:
> Do we want regular go apps dialing ws endpoints? Should we make protocol selection configurable?

I thought it would be good to have it exercised if someone has configured it, and it makes working with our quickstart single ER deployment a bit easier.  But I can pull it out when not building for js/wasm

edge-apis/component.go
```diff
@@ -27,10 +26,6 @@ func NewComponents(api *url.URL, schemes []string) *Components {
 
 	httpTransport := &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
-		DialContext: (&net.Dialer{
-			Timeout:   10 * time.Second,
-			KeepAlive: 10 * time.Second,
-		}).DialContext,
```
andrewpmartinez, line 33: I just re-ran my tests w/ this not set, and it worked just fine. I'm not sure why it failed the first time I tried it.

andrewpmartinez (approved):
lgtm

PR #444, opened 2023-10-31
Don't send close in response to close. Fixes #443

### LINKED ISSUE #443: Don't send close in reponse to a close on a listener

plorenz:


camotts (approved):

ziti/edge/network/listener.go
```diff
@@ -146,10 +150,7 @@ func (listener *edgeListener) Close() error {
 	edgeChan.hosting.Delete(listener.token)
 
 	defer func() {
-		if err := edgeChan.Close(); err != nil {
-			logger.WithError(err).Error("unable to close conn")
```
dovholuknf, line 150: this isn't possible to fail? i don't see why not log it. was it super chatty?

plorenz: Close() returns an error to conform to the io.Closer interface. However close(bool) doesn't since it doesn't need to conform to the interface, (and Close() always returned nil because it just called close()).

PR #451, opened 2023-11-07
Support idempotent terminator creation. Fixes #450

### LINKED ISSUE #450: Support idempotent terminator creation
Use consistent keys for a given service listener, so that repeated requests will work. Provide a listener id, so that the router can correlate requests.

plorenz:
Use consistent keys for a given service listener, so that repeated requests will work. Provide a listener id, so that the router can correlate requests.


ziti/edge/network/factory.go
```diff
@@ -76,7 +76,7 @@ func (conn *routerConn) BindChannel(binding channel.Binding) error {
 	return nil
 }
 
-func (conn *routerConn) NewConn(service *rest_model.ServiceDetail, connType ConnType) *edgeConn {
+func (conn *routerConn) NewDialConn(service *rest_model.ServiceDetail, connType ConnType) *edgeConn {
```
ekoby, line 79: seems that connType is redundant here

ziti/edge/network/factory.go
```diff
@@ -103,8 +103,28 @@ func (conn *routerConn) NewConn(service *rest_model.ServiceDetail, connType Conn
 	return edgeCh
 }
 
+func (conn *routerConn) NewListenConn(service *rest_model.ServiceDetail, connType ConnType, keyPair *kx.KeyPair) *edgeConn {
```
ekoby, line 106: connType is not needed

ziti/edge/network/factory.go
plorenz: I think it used to be one method, and I didn't notice this when I split it into two. Fixed.

ziti/edge/network/factory.go
plorenz: Fixed

ekoby (approved):

PR #463, opened 2023-12-11
Allow refreshing a single service. Fixes #462

### LINKED ISSUE #462: Allow refreshing a single service
Allow refreshing a single service, rather than doing a full service refresh, for those cases where you know what service you need information about, which may have just been added.

plorenz:


michaelquigley (approved):
Seems reasonable to me? :shrug:

PR #469, opened 2023-12-14
Do fewer session refreshes. Fixes #468

### LINKED ISSUE #468: SDK does an unnecessary number of session refreshes
The SDK refreshes session every time we poll for services and before every dial. Edge sessions only need to be refreshed if there's a change to the edge router list for the session, which should be rare.

We should allow them to be refreshed on a separate configurable timer.

plorenz:


ziti/ziti.go
```diff
@@ -710,10 +710,29 @@ func (context *ContextImpl) RefreshService(serviceName string) (*rest_model.Serv
...
 	log := pfxlog.Logger()
-	svcUpdateTick := time.NewTicker(context.options.RefreshInterval)
-	defer svcUpdateTick.Stop()
+	svcRefreshInterval := context.options.RefreshInterval
+
+	if svcRefreshInterval == 0 {
+		svcRefreshInterval = DefaultServiceRefreshInterval
+	}
+	if svcRefreshInterval < time.Second {
+		svcRefreshInterval = time.Second
```
scareything, line 721: Maybe this is too aggressive for a default (when the caller probably made a mistake)? Also a `minRefreshInterval` constant to cover session and service refreshes might make sense.

scareything (approved):

ziti/ziti.go
plorenz: I was thinking that a developer might want to set it this low. So if the dev forgot to set it at all, it will get set to the default value, but if they have set it to something low, we'll cap it at 1 second.

I'll add a constant for minRefreshInterval

plorenz: constant added

PR #488, opened 2024-01-21
properly concatenate base url and path

### LINKED ISSUE #487: enrolling with 3rd party ca fails with expected 1 or more CAs from controller, got 0
see https://openziti.discourse.group/t/3rd-party-auto-enrollment/1993

**Steps to reproduce:**
Adding a 3rd party CA for auto-enrollment and then using the jwt from the CA results in "expected 1 or more CAs from controller, got 0".

**Expected Behavior:**
The identity should be enrolled properly and a new identity would exist in the controller

dovholuknf:
closes #487 

If the issuer in the jwt ends with a slash such as: `  "iss": "https://localhost:11280/",` the url to fetch the CA bundle would get concatenated with a doubled slash:

```
https://localhost:11280//edge/client/v1
```

### DIFF
--- a/edge-apis/urls.go
+++ b/edge-apis/urls.go
@@ -30,6 +30,9 @@ func ClientUrl(hostname string) string {
 		hostname = "https://" + hostname
 	}
 
+	if strings.HasSuffix(hostname, "/") {
+		return strings.Trim(hostname, "/") + ClientApiPath
+	}
 	return hostname + ClientApiPath
 }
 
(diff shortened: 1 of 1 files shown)


andrewpmartinez (changes requested):
Please mirror changes in `ManagementUrl()`

dovholuknf:
both are now trim/concatenated

andrewpmartinez (approved):

PR #490, opened 2024-01-22
Issue 489 log when no match

### LINKED ISSUE #489: no log when dialing a service that doesn't have matching intercept
if you create an identity using the intercept-based service location and neglect to give that identity rights to dial the service, nothing in the logs helps you figure this out. Please add a small log indicating that no match was found

dovholuknf:
closes #489 

Adds a log statement giving you a hint that your identity can't dial the requested address

### DIFF
--- a/ziti/dialer.go
+++ b/ziti/dialer.go
@@ -3,6 +3,7 @@ package ziti
 import (
 	"context"
 	"fmt"
+	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/edge-api/rest_model"
 	"math"
 	"net"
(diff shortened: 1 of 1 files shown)


dovholuknf:
I changed over to making an SdkCollection and adding a context to it, instead of using the deprectated `ziti.NewDialer()/ziti.NewDialerWithFallback` and a helpful error is output in that situation so I'm closing this PR/issue (for now at least)

You'll see something like:
```
"Failed to init cache: failed to send a http request -> Get \"http://core-metadata.edgex.ziti:80/api/v3/device/serv
ice/name/device-rest?limit=-1&offset=0\": address [tcp:core-metadata.edgex.ziti:80] is not intercepted by any ziti context"
```

PR #514, opened 2024-03-07
adds a temporary feature flag to enable OIDC (HA) based authentication

andrewpmartinez:
- existing implementations and configuration files will not have this value set and will default to false
- the default, false, state is to use legacy auth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+# Relase 0.23.1
+
+## What's New
+
+* `EnableHa` Feature Toggle
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
plorenz, line 1: 0.23.1 is already used, looks like

plorenz (approved):

CHANGELOG.md
andrewpmartinez: It was by ziti-ci. Looks like a dep update was done.

PR #558, opened 2024-04-29
fixes openziti/ziti#1980 updates for multiple controller options

andrewpmartinez:


ziti/enroll/enroll.go
```diff
@@ -299,11 +299,27 @@ func enrollUpdb(username, password string, token *ziti.EnrollmentClaims, caPool
...
 	}
 
-	resp, err := client.Post(token.EnrolmentUrl(), "application/json", bytes.NewBuffer(body.EncodeJSON()))
+	enrollmentUrls := token.EnrolmentUrls()
+
+	var resp *http.Response
+	var err error
+	for _, enrollmentUrl := range enrollmentUrls {
+		resp, err = client.Post(enrollmentUrl, "application/json", bytes.NewBuffer(body.EncodeJSON()))
+
```
plorenz, line 308: Do you want to break after the first non-erroring response?

andrewpmartinez: Why yes it should.

PR #594, opened 2024-07-16
fixes #593 wrong id value from api session

### LINKED ISSUE #593: SDK Golang OIDC Get API Session Returns Wrong Value
The `apiSession.GetId()` w/ an OIDC backed struct, returns the wrong ID value.

andrewpmartinez:


dovholuknf (approved):

plorenz:
If you run `ziti-ci update-sdk-build-info` it'll update the version number and fix the build

andrewpmartinez:
> If you run `ziti-ci update-sdk-build-info` it'll update the version number and fix the build

I feel like this should be something that is handled for us. I figured out what was wrong by searching the code base for the value it got in the error. I updated it by hand w/o knowing about that command.

PR #603, opened 2024-07-30
Only send config types on service list if controller version supports it. Fixes #601

### LINKED ISSUE #601: Only send config types on service list if controller version supports it
Only send config types on service list if controller version supports it. Requires minimum controller version 1.1.0

plorenz:


ziti/client.go
```diff
@@ -333,8 +338,9 @@ func (self *CtrlClient) GetServices() ([]*rest_model.ServiceDetail, error) {
 	for {
 		params.Limit = &pageLimit
 		params.Offset = &pageOffset
-		params.ConfigTypes = self.ConfigTypes
-
+		if self.supportsSetOfConfigTypesOnServiceList() {
```
dovholuknf, line 341: silly nit, but why are the Limit/Offset/ConfigTypes set in the loop. they never change, right? why not pull them outside the loop?

dovholuknf (approved):
perhaps a nit comment, but lgtm

ziti/client.go
plorenz: offset gets changed, but the other two don't. moved those out of the loop

PR #608, opened 2024-08-02
Update go-jose. Fixes #607

### LINKED ISSUE #607: go-jose v2.6.0 CVE-2024-28180 resolution
There is a CVE in go-jose v2.6.0
Our project depends on openziti sdk-golang, so this dependency is included
https://github.com/openziti/sdk-golang/blob/fb662f949bd0fbc6258f63ae5d963b2684f13958/go.mod#L89

Please see more details from the dependabot security adviosry
https://github.com/edgexfoundry/device-sdk-go/security/dependabot/11

The problem is fixed in the following packages and versions:
github.com/go-jose/go-jose/v4 version 4.0.1
github.com/go-jose/go-jose/v3 version 3.0.3
gopkg.in/go-jose/go-jose.v2 version 2.6.3

The problem will not be fixed in the following package because the package is archived:
gopkg.in/square/go-jose.v2

Are you able to resolve this dependency issue?

plorenz:


go.mod
```diff
@@ -86,6 +86,7 @@ require (
 	golang.org/x/sync v0.7.0 // indirect
 	golang.org/x/term v0.22.0 // indirect
 	golang.org/x/text v0.16.0 // indirect
+	gopkg.in/go-jose/go-jose.v2 v2.6.3 // indirect
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
```
andrewpmartinez, line 90: Two diff module paths `/square` vs `/go-jose` does that even fix the issue?

dovholuknf: FYI I pulled the branch and looked at the deptree from this project and the only go-jose tree I see is from zitadel so if 2.6.3 has the vulnerability fixed, it seems to me that this would resolve the issue here

plorenz:
@andrewpmartinez is working on getting the library updated in zitiadel.

PR #624, opened 2024-09-13
implement support for receiving multipart edge messages

ekoby:
this only implements support for receiving multipart messages

issue for sending multipart is #623

### DIFF
--- a/ziti/edge/conn.go
+++ b/ziti/edge/conn.go
@@ -153,13 +153,22 @@ func (ec *MsgChannel) WriteTraced(data []byte, msgUUID []byte, hdrs map[int32][]
 	copyBuf := make([]byte, len(data))
 	copy(copyBuf, data)
 
-	msg := NewDataMsg(ec.id, ec.msgIdSeq.Next(), copyBuf)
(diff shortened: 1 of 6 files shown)


ziti/edge/conn.go
```diff
@@ -153,14 +153,23 @@ func (ec *MsgChannel) WriteTraced(data []byte, msgUUID []byte, hdrs map[int32][]
...
-	msg := NewDataMsg(ec.id, ec.msgIdSeq.Next(), copyBuf)
+	seq := ec.msgIdSeq.Next()
+	msg := NewDataMsg(ec.id, seq, copyBuf)
 	if msgUUID != nil {
 		msg.Headers[UUIDHeader] = msgUUID
 	}
 
 	for k, v := range hdrs {
 		msg.Headers[k] = v
 	}
```
plorenz, line 164: Could we send this back on the connect reply instead?

ekoby: connect reply goes to ER/controller and not go the peer SDK

plorenz (approved):

PR #658, opened 2025-01-13
feat: update edge management examples to reflect current state of the…

nenkoru:
While working with the sdk I found out that the current way to instantiate an Edge Management API client differs from the one in the example.
Hence the proposed changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,10 @@ should not be necessary. This SDK provides a wrapper around the generated client
 
 #### Example: Creating an Edge Management API Client
 ```golang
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -324,7 +331,7 @@ credentials.CaPool = caPool
 
 //Note: the CA pool can be provided here or during the Authenticate(<creds>) call. It is allowed here to enable
 //      calls to REST API endpoints that do not require authentication.
```
plorenz, line 333: This should also take the totpCallback, or nil.

PR #662, opened 2025-01-22
tweak NewSdkCollectionFromEnv and fix the http-client example

dovholuknf:


ziti/collection.go
```diff
@@ -62,7 +62,7 @@ func NewSdkCollection() *CtxCollection {
 
 // NewSdkCollectionFromEnv will create an empty CtxCollection and then attempt to populate it from configuration files
 // provided in a semicolon separate list of file paths retrieved from an environment variable.
-func NewSdkCollectionFromEnv(envVariable string) *CtxCollection {
+func NewSdkCollectionFromEnv(envVariable string, configTypes []string) *CtxCollection {
```
andrewpmartinez, line 65: This is an externally consumed function, this would be a major bump if left as as.

You should make it take an an optional number of strings, which can be 0 or more.

`NewSdkCollectionFromEnv(envVariable string, configTypes ...string)`

Which allows 0 or more config types to be specified. You still access configTypes like a slice.

dovholuknf: that is a grand idea. done

plorenz (approved):

PR #674, opened 2025-01-31
Add license check. Fixes #673

### LINKED ISSUE #673: Add license check to GH workflow

plorenz:


andrewpmartinez (approved):
This is ...fine. Ideally, it would have been nice to have a whitelist/blacklist. But that might be hard to maintain.

I assume the idea here is that if the licenses change in any way, someone has to go and add their licenses to the expected list, which hopefully get flagged in a PR.

plorenz:
> This is ...fine. Ideally, it would have been nice to have a whitelist/blacklist. But that might be hard to maintain.
> 
> I assume the idea here is that if the licenses change in any way, someone has to go and add their licenses to the expected list, which hopefully get flagged in a PR.

Yes, exactly. I wouldn't expect that to happen often, since most projects should fall under the set already tracked.

PR #705, opened 2025-03-30
Replace `golang.org/x/exp` with stdlib

Juneezee:
These experimental packages are now available in the Go standard library.

1. `golang.org/x/exp/slices` -> `slices` (https://go.dev/doc/go1.21#slices)
2. `golang.org/x/exp/rand` -> `math/rand/v2` (https://go.dev/doc/go1.22#math_rand_v2)

`golang.org/x/exp/rand` has been deprecated and scheduled to be deleted.

Reference: https://github.com/golang/exp/commit/f9890c6ad9f380fd3cdb66a33843f522d4790e03


plorenz (approved):
Looks good. Thank you for contributing! I'll merge this before doing the next release.

PR #718, opened 2025-04-23
Fix ER connection race condition leak. Fixes #717

### LINKED ISSUE #717: ER connection race condition can leak connections
If two connections to an ER are made at the same time, the duplicate gets removed, but we don't check that the entry removed from cache is the duplicate, so both can get removed, but only the duplicate gets closed.

plorenz:


ziti/ziti.go
```diff
@@ -486,8 +486,13 @@ func (context *ContextImpl) Sessions() ([]*rest_model.SessionDetail, error) {
...
 	logrus.Debugf("connection to router [%s] was closed", routerConn.Key())
-	context.Emit(EventRouterDisconnected, routerConn.GetRouterName(), routerConn.Key())
-	context.routerConnections.Remove(routerConn.Key())
+	removed := context.routerConnections.RemoveCb(routerConn.Key(), func(key string, v edge.RouterConn, exists bool) bool {
+		return exists && v == routerConn
+	})
+
+	if removed {
+		context.Emit(EventRouterDisconnected, routerConn.GetRouterName(), routerConn.Key())
+	}
```
andrewpmartinez, line 495: Removing this is ...ok?

plorenz: If we create two connections to the router at the same time, and one is a duplicate, we close the duplicate. Should we emit a router disconnect if we close a duplicate connection, or only if we close the connection that's actually being used?

andrewpmartinez: I don't have a clear answer. So here are some thoughts I have:

In a perfect world, we would go through some connection coordination pool/requester/etc. to ERs that would handle that for us, so we don't have a race condition. I don't recall enough about how these systems interact to know if that's even a reasonable goal to pursue.

Barring that, if I am in a world with multiple outgoing connections, if we have an existing connection that is serviceable and we don't close/lose it and continue to use it, then not emitting a connect/close is fine. I'm not sure if any tests would react to that - my gut says no.

plorenz: Ah, I missed the connect event. I moved that so that it only gets emitted if we use the new connection. Agree that we should ideally only make one connection at a time, but I don't want to get distracted by refactoring that at the moment.

michaelquigley: seems reasonable to me? :+1:

michaelquigley (approved):

PR #747, opened 2025-06-18
Adding router proxy support

cx-serhii-stanislavskyi:
# Add Router Proxy Configuration from Environment Variables

This PR adds two new functions to improve router connectivity options:

## Changes

1. **Added `routerProxyFromEnvironment` function**
   - Enables automatic proxy configuration for router connections based on environment variables
   - Uses `HTTPS_PROXY` or `HTTP_PROXY` environment variables to determine proxy settings
   - Automatically applied when loading configuration from file via `NewConfigFromFile()`

2. **Added `parseTLS(raw string)` helper function**
   - Supports router URLs with the format `tls:router.example.com:443`
   - Converts `tls:` scheme URLs to proper HTTPS URLs for processing
   - Enables more flexible router addressing options

## Use Cases

- Simplifies deployment in environments where direct connections are restricted and proxies are required
- Allows for consistent proxy configuration across applications using environment variables
- Supports more intuitive TLS connection specification with the `tls:` URL format

These changes maintain backward compatibility while adding flexibility for different network configurations.

### DIFF
--- a/ziti/config.go
+++ b/ziti/config.go
@@ -121,7 +121,8 @@ func NewConfigFromFile(confFile string) (*Config, error) {
 		// Parse the HTTPS_PROXY env (or https:// proxy setting) for this address
 		req := &http.Request{URL: &url.URL{Host: addr}}
 		proxyURL, errProxy := http.ProxyFromEnvironment(req)
-		pfxlog.Logger().Infof("!!!!!!!!url: %s, proxyURL: %v, errProxy: %v", addr, proxyURL, errProxy)
+		val := os.Getenv("HTTPS_PROXY") // for debugging purposes
+		pfxlog.Logger().Infof("!!!!!!!!val: %s, url: %s, proxyURL: %v, errProxy: %v", val, addr, proxyURL, errProxy)
 		if proxyURL == nil {
 			return nil // no proxy
 		}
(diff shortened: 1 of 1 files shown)


ziti/config.go
```diff
@@ -127,3 +130,44 @@ func NewConfigFromFile(confFile string) (*Config, error) {
 func GetControllerWellKnownCaPool(controllerAddr string) (*x509.CertPool, error) {
 	return rest_util.GetControllerWellKnownCaPool(controllerAddr)
 }
```
plorenz, line 132: Apologies for the late review. 

First, if you haven't yet, please take a look at https://github.com/openziti/sdk-golang/blob/main/CONTRIBUTING.md as a CLA is required for contributors.

Second, overall looks good. I'd like this function to be available to SDK users, but not have it set on RouterProxy by default. To that end, could you remove the set and make the method public?

cx-serhii-stanislavskyi: My apologies for the late response. 
1) We're still clarifying question with CLA with legal team.
2) I made it method public but for me is tricky how to implement it in zrok. 
We're expecting it routerProxy configuration here:
https://github.com/openziti/sdk-golang/blob/main/ziti/ziti.go#L1461
`if context.routerProxy != nil {
		if proxyConfig := context.routerProxy(ingressUrl); proxyConfig != nil {
			dialerConfig.TransportConfig[transport.KeyCachedProxyConfiguration] = proxyConfig
		}
	}` 
That's inside connectEdgeRouter method. 
It's for passing proxy configuration to edge router when we're creating public share

PR #778, opened 2025-07-25
Reauthenticate if OIDC refresh fails. Fixes #777

### LINKED ISSUE #777: OIDC auth token refresh doesn't fall back to re-auth if token has expired
Manifests as
```
[298180.888]    INFO sdk-golang/ziti.(*ContextImpl).RefreshApiSessionWithBackoff.func1: {error=[ErrorType=invalid_request Description=subject_token is invalid]} unable to refresh apiSession, error type *oidc.Error, will retry
```
Happens if the system is put to sleep long enough for the JWT to expire.

plorenz:


plorenz:
> Do we need to match expiry specifically, or is it sufficient to re-auth on any oidc error?

Looking at the error types that can be returned on OIDC error, they generally seemed unrecoverable, so fallback to full auth seemed to make the most sense. I've ping Andrew about this, so I may have a follow-up PR if he disagrees.

PR #797, opened 2025-09-02
Add.generic.msg.sink

andrewpmartinez:
Should this have a major version bump?


ziti/edge/network/conn.go
```diff
@@ -51,7 +51,7 @@ type edgeConn struct {
 	edge.MsgChannel
 	readQ                 *noopSeq[*channel.Message]
 	inBuffer              [][]byte
-	msgMux                edge.MsgMux
+	msgMux                edge.ConnMux[any]
```
andrewpmartinez, line 54: Nothing in the SDK is currently using the generic data for each conn, so I put `any` in. However, IIRC there was a future use case for this to add context data for reporting/debugging.

PR #812, opened 2025-10-01
fixes #811 credentials are lost when set

### LINKED ISSUE #811: Credentials are lost when explicitly set
*Experienced*

If credentials in a ziti context config are explicitly set, they are lost.

*Expected*

Credentials are used if set. If not set, default to checking for an identity config.

*Reproduction*

1) Create some credentials `certCreds := &edgeapis.CertCredentials{...}`
2) Set them on a ztx config: `ziti.Conf{Credentials: certCreds}`
3) Attempt to authenticate `ztx.Authenticate()`

andrewpmartinez:
- adds test for id config vs cred setting

### DIFF
--- a/ziti/contexts.go
+++ b/ziti/contexts.go
@@ -98,9 +98,11 @@ func NewContextWithOpts(cfg *Config, options *Options) (Context, error) {
 		newContext.maxDefaultConnections = 1
 	}
 
-	idCredentials := edge_apis.NewIdentityCredentialsFromConfig(cfg.ID)
-	idCredentials.ConfigTypes = cfg.ConfigTypes
-	cfg.Credentials = idCredentials
+	if cfg.Credentials == nil {
+		idCredentials := edge_apis.NewIdentityCredentialsFromConfig(cfg.ID)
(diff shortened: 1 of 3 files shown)


ziti/contexts.go
```diff
@@ -98,9 +98,11 @@ func NewContextWithOpts(cfg *Config, options *Options) (Context, error) {
...
 	}
 
-	idCredentials := edge_apis.NewIdentityCredentialsFromConfig(cfg.ID)
-	idCredentials.ConfigTypes = cfg.ConfigTypes
-	cfg.Credentials = idCredentials
+	if cfg.Credentials == nil {
+		idCredentials := edge_apis.NewIdentityCredentialsFromConfig(cfg.ID)
+		idCredentials.ConfigTypes = cfg.ConfigTypes
+		cfg.Credentials = idCredentials
+	}
```
dovholuknf, line 105: would you consider adding a warning in the else here? this might be confusing to people imo. Or return an error ?

dovholuknf (approved):

PR #816, opened 2025-10-06
adds router posture responses (ha support)

andrewpmartinez:
- adds ability to submit posture to controller or router based on router capabilities
- adds new ToTP MFA JWT token type for MFA posture checks post-login

### DIFF
--- a/ziti/edge/posture/cache.go
+++ b/ziti/edge/posture/cache.go
@@ -150,7 +150,9 @@ func (cache *Cache) Evaluate() {
 	activeDialServices := cache.serviceProvider.GetActiveDialServices()
 	activeBindServices := cache.serviceProvider.GetActiveBindServices()
 
-	activeQueryTypes, activeProcesses := getActiveQueryInfo(activeDialServices, activeBindServices)
+	activeQueryTypes, activeProcesses, lowestTotpTimeout := getActiveQueryInfo(activeDialServices, activeBindServices)
+
+	cache.setTotpTimeout(lowestTotpTimeout)
 
 	candidateData := NewCacheData()
 	candidateData.Index = cache.currentData.Index + 1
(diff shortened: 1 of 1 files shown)


pb/edge_client_pb/edge_client.proto
```diff
@@ -135,8 +136,13 @@ message PostureResponse {
     Woken woken = 5;
     Unlocked unlocked = 6;
     SdkInfo sdkInfo = 7;
+    TotpToken totpToken = 8;
   };
```
plorenz, line 141: What's the advantage of wrapping this in a type, instead of just using string?

ziti/edge/messages.go
```diff
@@ -350,6 +358,125 @@ func NewDialFailedMsg(connId uint32, message string) *channel.Message {
...
+
+// PostureResponseTotp is a posture response that contains a TOTP token. Used to track the last time
+// a client has performed TOTP code submission. This exists outside of the REST APIs as it is only
+// needed in HA systems and is not used for legacy API Sessions. It does conform to the REST APIs generated
+// interfaces in order for it to be included in interface typed slices.
+type PostureResponseTotp struct {
+	TotpToken string //should be a signed, API Session scoped JWT. See common.TotpClaims
+	Id        *string
+}
+
```
plorenz, line 457: Should probably be using pointer receivers, see staticcheck error below

ziti/edge/network/factory.go
```diff
@@ -128,6 +128,21 @@ func (conn *routerConn) NewDialConn(service *rest_model.ServiceDetail) *edgeConn
 	return edgeCh
 }
```
plorenz, line 130: Would it make sense to have a container message if you're sending multiple responses?

andrewpmartinez: I actually do have it already. I just failed to use it.

ziti/edge/messages.go
andrewpmartinez: You are correct. Fixed.

pb/edge_client_pb/edge_client.proto
andrewpmartinez: I was unsure if other fields would be necessary in the future. I could use a string here, but if a field is added later, it would be backward-incompatible or I would have to add a different but similarly named field.

PR #864, opened 2026-02-10
for openziti/ziti#3356 adds ability to see OIDC responses

andrewpmartinez:
- adds optional AuthenticateWithResponses() to inspect all OIDC flow requests


edge-apis/clients_shared.go
```diff
@@ -68,6 +67,22 @@ type OidcEnabledApi interface {
 	SetOidcRedirectUri(redirectUri string)
 }
```
plorenz: duing -> during

PR #869, opened 2026-02-13
Add support for using contexts for timeout/cancellation on dial. Fixes #708

### LINKED ISSUE #708: Support for Go's built-in context in Dial methods
Currently ziti.Context's Dial/DialWithOptions/DialAddr does not support passing golang's built-in context.Context object for cancellation
While it does support passing a ConnectTimeout, I'd like to be able to cancel the Dial  attempt on-demand and not wait for a timeout

plorenz:


ekoby (approved):
looks good, other than normalizing default timeout

Issue #901, opened 2026-03-21
Move xgress back to having retransmitter goroutine per-xgress


plorenz:
Allows the re-transmitter to block on sends, allowing it to properly respond to back-pressure. It also simplifies the code.

PR #904, opened 2026-03-21
Move xgress back to having retransmitter goroutine per-xgress. Fixes …

plorenz:
…#901


xgress/retransmitter.go
```diff
@@ -1,182 +0,0 @@
...
-		faultReporter:    faultReporter,
-
-		retransmissions:        metrics.Meter("xgress.retransmissions"),
-		retransmissionFailures: metrics.Meter("xgress.retransmission_failures"),
-	}
-
-	go ctrl.retransmitIngester()
-	go ctrl.retransmitSender()
-
-	metrics.FuncGauge("xgress.retransmits.queue_size", func() int64 {
```
andrewpmartinez, line 56: This one seems to be gone in the new code. Not sure how important it was.

xgress/retransmitter.go
```diff
@@ -1,182 +0,0 @@
...
-		select {
-		case retransmit := <-self.retransmitSend:
-			if !retransmit.isAcked() {
-				retransmit.payload.MarkAsRetransmit()
-				if err := retransmit.x.dataPlane.RetransmitPayload(retransmit.x.address, retransmit.payload); err != nil {
-					// if xgress payload buffer is closed, don't log the error
-					if !retransmit.x.payloadBuffer.IsClosed() {
-						logger.WithError(err).Errorf("unexpected error while retransmitting payload from [@/%v]", retransmit.x.address)
-						self.retransmissionFailures.Mark(1)
-						self.faultReporter.ReportForwardingFault(retransmit.payload.CircuitId, retransmit.x.ctrlId)
```
andrewpmartinez, line 168: I assume forwarding faults are handled differently now? I'm not sure whether this was used by routers or something else.

andrewpmartinez (approved):

xgress/retransmitter.go
plorenz: The queue is gone, since the rtxer is no longer global. We might be able to introduce a global pending retransmits with a shared counter

xgress/retransmitter.go
plorenz: Good catch. I'll need to add the fault notification on the router side in the RetransmitPayload method

PR #907, opened 2026-03-30
Fix potential nil references on session service structs. Fixes #906

### LINKED ISSUE #906: Fix potential nil references on session service structs

plorenz:


plorenz:
> So we don't need to check if session.ServiceID is nil?
> 
> I assume this is a lookup issue with the session.Service didn't get set. Howver seems odd to move from one blind deref to another.

A session with a nil ServiceID is a programming error. However, there are cases on refresh where the service struct doesn't get rebuilt and re-assigned. Could argue that that is also a programming error, but I'm not sure the struct is necessary

PR #928, opened 2026-04-29
auth fixes

### LINKED ISSUE #924: Make controller http timeout configurable, with a default of 30s

### LINKED ISSUE #925: Switch controllers on a broader set of errors
Switch controllers on 
* HTTP timeouts
* context cancellation
* EOF / ErrUnexpectedEOF
* 5xx responses
* url.Error timeouts

### LINKED ISSUE #926: Refresh OIDC token using a window to avoid race conditions and herding
* replaces the fixed expiresAt - 10s schedule with picking a random point between 1/2 and 5/6 of the token's remaining lifetime. Avoids problems with slow refreshes or temporary controller unavailability causing our only attempt, right at the end of the token lifetime to fail. Also, spreading refreshes avoid herding issues.
* skip the periodic service-list refresh when a token refresh is due inunder 15 seconds, so the service refresh doesn't race the expiring token into an unnecessary re-auth on a 401.

### LINKED ISSUE #927: Apply exponential backoff to auth retry attempts

plorenz:
- **Authenticate ListServices call in CtrlClient.GetService**
- **Make controller HTTP timeout configurable, raise default to 30s. Fixes #924**
- **Broaden controller-swap error matching, fix shared errors.As target bug. Fixes #925**
- **Refresh access tokens using a C-SDK style window. Fixes #926**
- **Apply exponential backoff to failed authentication attempts. Fixes #927**
- **Skip re-authentication on 401 within the refresh race window**
- **Update deps and changelog**


ziti/ziti.go
```diff
@@ -1198,21 +1265,75 @@ func (context *ContextImpl) Authenticate() error {
...
-		if time.Since(context.lastSuccessfulApiSessionRefresh) < 5*time.Second {
+		if time.Since(context.lastSuccessfulApiSessionRefresh.Load()) < 5*time.Second {
+			context.resetAuthBackoff()
 			return nil
 		}
+	}
+
+	if !context.lastAuthAttempt.IsZero() && context.currentWait > 0 {
+		if remaining := context.currentWait - time.Since(context.lastAuthAttempt); remaining > 0 {
+			logrus.Debugf("auth backoff: %v remaining, returning cached error", remaining)
```
andrewpmartinez: no pfxlog?

andrewpmartinez:
Sent some private comments.

andrewpmartinez (changes requested):

ziti/ziti.go
```diff
@@ -868,6 +873,20 @@ func (context *ContextImpl) RefreshServices() error {
...
+// refreshRaceWindow is how long after a successful api-session refresh the
+// SDK treats a 401 on a service-related call as a propagation/in-flight
+// race rather than a real auth failure. Within this window the 401 handler
+// simply retries with the current (already-refreshed) session instead of
+// invoking Authenticate, which would needlessly drive the refresh machinery
+// and may churn sessions/ER connections if the refresh path decides a full
+// re-auth is required.
+const refreshRaceWindow = 15 * time.Second
+
+func recentlyAuthenticated(context *ContextImpl) bool {
```
andrewpmartinez, line 884: probably should have some godoc here considering how complex this logic is.

plorenz: the doc is above, on refreshRaceWindow

ziti/ziti.go
plorenz: fixed

plorenz:
Review comments addressed

andrewpmartinez (approved):

PR #937, opened 2026-05-28
changelog updates

andrewpmartinez:


plorenz (approved):
Looks good, as long as we're updating the correct version

PR #943, opened 2026-06-06
Connect-V2, SDK acceptance-test framework, and the 2.0/v2 migration

### LINKED ISSUE #936: Implement Connect-V2: sessionless SDK dial
Implements the Connect-V2 dial protocol — sessionless SDK dials that
authorize at the router via RDM instead of requiring a controller-issued
service session token.

## Why

- Removes a controller round-trip per dial.
- Aligns the SDK with the router's local RDM-based authorization.
- Lays groundwork for SDK-hosted xgress and direct SDK-to-SDK paths.

## What's included (this repo)

- Sessionless dial path: `DialContextWithOptions` no longer creates a
  service session up front. Edge routers are discovered via the
  existing `GET /edge/client/v1/services/{id}/edge-routers` endpoint
  through a new `CtrlClient.GetServiceEdgeRouters` wrapper.
- Per-service ER cache on `ContextImpl.serviceEdgeRouters`, refreshed
  on the same cadence as `refreshSessions` and warming router
  connections the same way. Cache invalidated on service removal and
  on dial failure.
- `dialSession` refactored: V2 path is sessionless on the wire; V1 path
  remains and now creates the service session lazily, only when V1 is
  actually taken.
- Force `UseXgressToSdkHeader=true` on Connect-V2 dials in the go-SDK
  (returned conn is always `edgeConnXgress`). Router-side continues to
  support both flow-control modes for other SDKs.
- User-facing `DialOptions.ForceConnectV1` escape hatch routes through
  the V1 path even when the router advertises V2.
- Drop `route_circuit` / `pending_dials` — the xgress `CircuitStart`
  handshake already serializes data flow, so the race window the
  module was guarding against doesn't exist. Circuit ID is read from
  `state_connected`'s `CircuitIdHeader`.
- Minimum controller/router version enforced at 1.0.0
  (`CtrlClient.requireMinControllerVersion`). Dev builds reporting
  `0.0.0` are accepted. Panics with a clear message on older
  controllers. Breaking change documented in `CHANGELOG.md`.
- Pre-existing fixes uncovered during the V2 audit:
  - `splitMultipart` bounds-checks length prefixes and returns
    descriptive errors instead of panicking on truncated input.
  - `handlePayloadWithNoSink` now sends the constructed ack instead of
    the original message.
  - `edgeConnXgress.close()` no longer NPEs when invoked from
    `setupXgressFlowControl`'s header-validation error paths before
    `xg` / `writeAdapter` are populated.

## Coordinated with

openziti/ziti#3884 — the matching controller/router counterpart.

### LINKED ISSUE #951: Add an SDK acceptance-test framework
## What

A general-purpose acceptance-test framework that exercises the Go SDK against real, versioned OpenZiti controllers and routers rather than mocks. It runs the SDK under test against a version matrix:

- `maint-lts` / `active-lts` / `latest` release lines (downloaded from GitHub releases)
- `main` and arbitrary git refs (branch/tag/SHA), built from source
- an optional co-development mode that builds a ziti ref against the local SDK tree

## Why

To catch SDK / controller-router compatibility regressions across versions, and to guarantee connect-v2 (sessionless V2 + OIDC) is genuinely exercised in CI rather than adaptively downgraded to a V1 pass.

## How

- the versioned `ziti` binary runs out of process; a quickstart controller and edge routers run as child processes
- the SDK under test is wired in via `replace => ../`
- admin setup is CLI-driven; tests follow an isolation contract (uniquely named resources, targeted policies, serial by default)
- a GitHub Actions matrix runs the suite per version, plus a dedicated required-mode connect-v2 job that fails rather than passing as V1

Design and decisions: `acceptance-tests.md`.

## Status

Implemented alongside the connect-v2 work (harness, P0 suite, version acquisition and source builds, CI workflow). Follow-up: the P1/P2 test expansion (revocation, posture, token refresh, UPDB, connection semantics, intercept dial, ...).

### LINKED ISSUE #952: xgress client half-close not delivered to legacy edge hosts
## Symptom

A ConnectV2 (xgress) client that half-closes its send side (`CloseWrite`) to a host bound *without* SDK xgress (the router terminates xgress and bridges to a legacy edge host) never delivers the half-close. A host that reads to EOF starves, and the round trip hangs.

## Root cause

Moving half-close into xgress replaced the edge FIN mechanism with the native xgress EOF flag. Historically half-close rode as an edge FIN carried transparently as a payload header that the terminating router maps onto `edge.FlagsHeader`. The native EOF flag is consumed by the xgress framework and never reaches a peer that does not speak it, so an older router bridging to a legacy host (or an older SDK) never sees the half-close. This is a back-compat break introduced by an SDK change.

## Fix

When the peer has not negotiated native EOF, `edgeConnXgress.CloseWrite` emits the legacy edge FIN as a payload header instead of the native flag; the terminating router already maps that header back to `edge.FlagsHeader`, so the host sees an ordinary FIN. The receive side already handled the inbound header. No router change is required, and it supports older routers.

## Verification

Red/green acceptance regression `Test_HalfClose_XgressClientToLegacyHost`: an xgress client half-closing to a router-bridged legacy host. Without the fix it hangs to the read deadline; with the fix it round-trips.

Addressed alongside the connect-v2 work.

plorenz:
Implements the Connect-V2 sessionless SDK dial protocol — dials are authorized at the router via RDM instead of requiring a controller-issued service session token.

- Adds `CtrlClient.GetServiceEdgeRouters` wrapping the existing `GET /edge/client/v1/services/{id}/edge-routers` endpoint for sessionless edge-router discovery.
- Adds a per-service ER cache on `ContextImpl.serviceEdgeRouters`, refreshed on the `sessionRefreshTimer` cadence and warming router connections the same way the session cache does. Invalidates on service removal and on dial failure.
- Refactors `dialSession` to skip service-session creation on the V2 path; V1 fallback now creates the service session lazily, only when the V1 wire is actually taken.
- Forces `UseXgressToSdkHeader=true` on Connect-V2 dials (the go-SDK V2 path is always `edgeConnXgress`); the router continues to support both flow-control modes for other SDKs.
- Adds `DialOptions.ForceConnectV1` escape hatch to route through the V1 path even when the router advertises V2.
- Removes `route_circuit` / `pending_dials`. The xgress `CircuitStart` handshake already serializes data flow, so the race window the module was guarding against does not exist. Reads the circuit ID from `state_connected`'s `CircuitIdHeader` instead.
- Uses the sessionless ER-list endpoint on controllers 1.0.0 and newer (`0.0.0` dev builds included), gated by `CtrlClient.supportsServiceEdgeRouterList`. Older controllers don't expose the endpoint on the client API; for those the SDK falls back to creating a dial session and using its attached edge routers, with the session cache as the source of truth. While the controller version is unknown, dials take the V1 path and the version load is retried on the next capability check. If the sessionless endpoint errors despite the version saying it should exist, the dial falls back to the V1 session path at runtime and drops the cached ER list. Documents the supported controller set (the 1.6.x and 2.0.0 LTS releases) in CHANGELOG and bumps the version from 1.7 to 2.0.
- Creates fallback sessions on the dial path via `createSessionWithBackoff`, so foreground dials get ctx cancellation, retry with backoff, re-auth on 401 and service-recreation handling; the background refresh loop uses a plain cached lookup.
- Restructures `createSessionWithBackoff` to run the retry before returning the session, removing a dead pre-retry cache call and a return statement that relied on operand evaluation order.
- Removes `GetRouterId()` from `edge.Conn` and `edge.MsgChannel`; documents the removal as a breaking change in CHANGELOG.
- Bounds-checks `splitMultipart` length prefixes; returns descriptive errors instead of panicking on truncated input.
- Fixes `handlePayloadWithNoSink` to send the constructed ack instead of the original message.
- Removes the `conn.Close()` calls from `setupXgressFlowControl`'s header-validation error paths so the conn no longer NPEs before `xg` / `writeAdapter` are populated.

For openziti/ziti#3884.

## Also in this PR (the 2.0 release)

This PR is now the 2.0 release. On top of Connect-V2 it lands the SDK acceptance-test framework, the xgress half-close back-compat fix, and the `/v2` module migration.

- **SDK acceptance-test framework** — exercises the SDK against real, versioned controllers/routers (release downloads and source builds), with a GitHub Actions matrix and a dedicated required-mode Connect-V2 coverage job. Design and decisions in `acceptance-tests.md`. Fixes #951.
- **Half-close back-compat** — an xgress (Connect-V2) client half-closing to a router-bridged legacy edge host now emits the legacy edge FIN when the peer hasn't negotiated native xgress EOF, restoring half-close delivery to hosts that read to EOF. Covered by a red/green acceptance regression. Fixes #952.
- **2.0 / v2 module path** — the module is now `github.com/openziti/sdk-golang/v2`; consumers update their imports to the `/v2` form. See CHANGELOG.

Note: the dedicated Connect-V2 CI job builds the ziti `connect-v2` branch against this SDK, so it needs that branch to also import `sdk-golang/v2`.

Fixes #936
Fixes #951
Fixes #952


ziti/edge/conn.go
```diff
@@ -123,7 +154,6 @@ type ServiceConn interface {
 type Conn interface {
 	ServiceConn
 	Identifiable
-	GetRouterId() string
```
andrewpmartinez, line 126: This removal is a potentially breaking change for external users and denotes a need for v2.0.0

andrewpmartinez:
```
conn, _ := zitiCtx.Dial("my-service")
routerId := conn.GetRouterId()      
```

ziti/client.go
```diff
@@ -529,29 +530,39 @@ func (self *CtrlClient) RemoveMfa(code string) error {
...
 func (self *CtrlClient) loadCtrlCapabilities() {
 	result, _ := self.API.Informational.ListVersion(informational.NewListVersionParams())
 	if result != nil && result.Payload != nil && result.Payload.Data != nil {
 		if sv, err := versions.ParseSemVer(result.Payload.Data.Version); err == nil {
+			self.controllerVersion.Store(sv)
 			if sv.Equals(versions.MustParseSemVer("0.0.0")) || sv.CompareTo(versions.MustParseSemVer("1.1.0")) >= 0 {
 				self.supportsConfigTypesOnServiceList.Store(true)
 			}
+			self.capabilitiesLoaded.Store(true)
 		}
```
andrewpmartinez, line 588: If, for some reason, this doesn't parse, it will run on every Dial/Connect.

If we can't parse the version/capabilities, we might be in a dead-in-the-water situation.

ziti/client.go
```diff
@@ -561,6 +572,42 @@ func (self *CtrlClient) supportsSetOfConfigTypesOnServiceList() bool {
...
+
+	resp, err := self.API.Service.ListServiceEdgeRouters(params, self.GetCurrentApiSession())
+	if err != nil {
+		return nil, rest_util.WrapErr(err)
+	}
+
+	ers := resp.Payload.Data.EdgeRouters
+	for _, er := range ers {
+		self.sanitizeEdgeRouterUrls(er)
+	}
```
andrewpmartinez, line 631: Should probably nil check resp.Payload/Data

version
```diff
@@ -1 +1 @@
-1.8
```
andrewpmartinez: I am confused now. 1.9 here but 2.0 in the changelog?

andrewpmartinez (dismissed):
The 2.0 changelog, vs the 1.9 version file, vs the breaking changes need to be resolved.

ziti/edge/conn.go
plorenz: We're now at /v2

ziti/client.go
plorenz: fixed

ziti/client.go
plorenz: added a throttle so we only check every 30s at most

version
plorenz: fixed, it's all v2.0.0 now

andrewpmartinez (approved):

PR #960, opened 2026-06-24
Add acceptance test for dialer connection metadata

plorenz:
Adds an acceptance test asserting that a host's accepted connection sees the dialer-supplied metadata: the caller id (the dialing identity's name), the app data passed on dial, and the dialer identity (asserted only on controller versions that populate it). The metadata assertions ride on a verified data plane: write, half-close, and read-to-EOF must succeed first.

One of six independent SDK acceptance-test PRs; each targets main.


plorenz:

PR #961, opened 2026-06-24
Add acceptance test for UPDB (username/password) auth

plorenz:
Adds an acceptance test for username/password (UPDB) login through the SDK, plus an api-session token refresh: a UPDB identity dials a service, then refreshes its session and dials again.

Harness: helpers to create and enroll a UPDB identity and build a UPDB-authenticated context.


plorenz:

PR #962, opened 2026-06-24
Add acceptance test for service-access update propagation

plorenz:
Adds an acceptance test that the SDK reflects service-access edits made after an initial dial: the changed/removed/added service events fire on a refresh, and cached state is invalidated rather than dialing stale data (a dial fails once access is revoked and succeeds again once re-granted).

Harness: a deletable policy handle so a test can revoke access mid-run.


plorenz:

PR #963, opened 2026-06-24
Add acceptance tests for address-based (intercept) dialing

plorenz:
Adds acceptance tests for the tunneler-style addressing entry point. With an intercept.v1 config, GetServiceForAddr scores an exact host:port match and DialAddr resolves the address and round-trips data. A service carrying only a ziti-tunneler-client.v1 config is still addressable, via the SDK's client-config-to-intercept conversion.

Harness: helpers to create configs, services that carry configs, and a context that requests given config types.


plorenz:

PR #964, opened 2026-06-24
Add acceptance tests for token refresh and hosting continuity

plorenz:
Adds acceptance tests that, across several api-session refreshes, a client keeps dialing (and on OIDC each refresh rotates the access token, which propagates to the edge routers), and that a hosted terminator survives the host's refresh and keeps accepting dials.

Independent; targets main.


plorenz:

PR #965, opened 2026-06-24
Add acceptance test for revocation enforcement

plorenz:
Adds an acceptance test that once a dialing identity is deleted, the SDK can no longer dial the service. Enforcement is asserted as eventual within a bounded window, since a freshly revoked identity may briefly dial on cached state. Reaper-driven teardown of an already-established circuit is left as a follow-up.

Independent; targets main.


plorenz:

PR #969, opened 2026-06-29
for openziti/ziti#3990 subscribe to router-pushed service and posture…

andrewpmartinez:
… changes

    - adds ServiceChangeSet and PostureStateChange protobuf messages plus the Subscribe/Unsubscribe/ResyncPostureState controls: indexed atomic structural envelopes (services, policies, posture check definitions, identity-resolved config bodies) with a separate unindexed per-connection posture pass/fail stream
    - adds a subscription coordinator holding independent per-router state with index gap detection, duplicate drops, posture seq resync, and state-overlay pruning; the app-facing service cache materializes from the highest-index subscribed router and merges pushed fields into polled details
    - follows traffic with subscriptions: capable routers are selected via the router capability list, each router a dial or bind uses gets subscribed, idle subscriptions are swept with a floor of one retained, and controller polling pauses only while a confirmed push connection exists, remaining the backwards-compatibility fallback
    - submits posture data per router: each posture-capable connection receives only the responses its own pushed checks require, diffed against a per-connection baseline; pushed posture queries carry the MFA timeout for TOTP scheduling
    - exposes the per-router view publicly: RouterView (router id, name, address, push capability, subscription state, structural index) with services -> policies -> posture checks all carrying ids and the router's pushed pass/fail, via Context.GetRouterViews and the EventRouterViewChanged listener
    - returns typed ConnErrors from dial and bind (service and router ids, cause, failing posture check ids), supporting both errors.Is sentinels and errors.As extraction, mapped from the routers' structured denials
    - promotes SetErrorEventHandler and SetConnectionChangeHandler into edge.Listener; SessionListener shrinks to the legacy GetCurrentSession add-on
    - adds RouterConn.GetRouterId (the router's certificate identity) and moves the service-subscriptions capability onto the router capability bitmask
    - stops serializing the structured error's Cause, which broke unmarshaling on receipt and silently discarded denials

Depends on: 
- https://github.com/openziti/edge-api/pull/202
- https://github.com/openziti/edge-api/pull/204


pb/edge_client_pb/edge_client.proto
```diff
@@ -50,6 +50,17 @@ enum ContentType {
   PostureResponseType = 10800;
   ServicePostureStateRequestType = 108001;
   ServicePostureStateResponseType = 108002;
+
+  // SDK -> router: opt in to service change push notifications
+  SubscribeToServiceUpdatesType     = 10900;
```
ekoby, line 55: this block seems out or order -- should this be 109000?

ekoby (approved):

plorenz:
I added some capabilities checking in https://github.com/openziti/sdk-golang/pull/971 (still draft, I wanted to build the first parts of sdk rerouting first, to exercise it). 

We should make sure we're consistent with capabilities. Maybe we need an enum type that knows the bitmask and the string versions so we can check using the same variable against the EdgeRouter type from the controller and the mask from the router channel?

pb/edge_client_pb/edge_client.proto
andrewpmartinez: Mmmm, looks like the two above jumped up a power of 10.

andrewpmartinez: Also, the ones above are apparently not used. I'm not sure what they are for - leaving for now. Will address later.

andrewpmartinez:
> I added some capabilities checking in #971 (still draft, I wanted to build the first parts of sdk rerouting first, to exercise it).
> 
> We should make sure we're consistent with capabilities. Maybe we need an enum type that knows the bitmask and the string versions so we can check using the same variable against the EdgeRouter type from the controller and the mask from the router channel?

Unified the capabilities after talking about it w/ plorenz.

pb/edge_client_pb/edge_client.proto
ekoby, line 1: I really appreciate doc on the new message types/headers/etc. Can we add doc on the rest at some point?

edit: added #974

ekoby (approved):

PR #1005, opened 2026-08-24
Bound peer-declared sizes in the xgress decode and reassembly paths

### LINKED ISSUE #1004: UnmarshallPacketPayload indexes the buffer without bounds checks
`UnmarshallPacketPayload` (`xgress/xgress.go`) reads its buffer without checking
any length first, starting on its own first line:

```go
func UnmarshallPacketPayload(buf []byte) (*channel.Message, error) {
	flagsField := buf[0]
```

An empty buffer panics there. The rest of the function has the same problem
throughout:

- `buf[1]` and `buf[2:]` assume at least two bytes
- under `RttFlagMask`, `rest[0]` and `rest[1]` assume at least two more
- under `HeartbeatFlagMask`, `rest[len(rest)-8:]` computes a negative index when
  fewer than eight bytes remain
- `rest[:circuitIdSize]` trusts the size nibble from the wire against whatever
  is actually left
- `binary.Uvarint` can return a read count of zero (buffer too small) or a
  negative one (value overflows a uint64), and `rest[read:]` panics on the
  negative case

## Second defect, in readU8ToBytesMap

The nested `readU8ToBytesMap` looked guarded but is not, for a different reason.
`valSize` comes from `binary.Uvarint` as a `uint64`, and the bounds check
converts it:

```go
if len(buf) < int(valSize) {
    return nil, nil, fmt.Errorf(...)
}
result[k] = buf[:valSize]
```

Above `MaxInt64`, `int(valSize)` is negative, so `len(buf) < negative` is false,
the guard passes, and the slice expression panics with the original uint64. A
fuzz target found this in seconds:

```
slice bounds out of range [:13726120590279941950] with capacity 1
```

The comparison has to happen in `uint64` space: `valSize > uint64(len(buf))`.

## Reachability

This is registered as the packet producer for link underlays
(`channel.DatagramMessageStrategy(xgress.UnmarshallPacketPayload)` in ziti's
`router/xlink_transport`), and `channel` selects a datagram underlay when the
link endpoint type is `dtls`. On a datagram underlay `rxHello` is the same read
path as any other frame, so this decoder runs before the listener's
`ConnectionHandler` has verified anything. Unlike the rest of the message path,
it is not behind the bind handler.

In practice the dtls link transport is essentially unused, so the exposure is
theoretical. Stream (tls) links use `channel.ReadV2`, which is bounds-checked,
and never reach this function.

## Fix

Bounds-check each read and return an error instead of indexing. Add a fuzz
target over the entry point so the class is covered rather than the specific
inputs known today.

### LINKED ISSUE #1006: Chunked payload reassembly trusts the peer's declared size
`Xgress.nextTxPayload` reassembles chunked payloads using a total size the peer
declares as a leading uvarint. Three things in that block trust it.

## 1. The uvarint read count is not checked

```go
self.txChunkSize, payloadReadOffset = binary.Uvarint(payloadChunk.Data)
```

`binary.Uvarint` returns `0` when the buffer is too small and a **negative**
count when the value overflows a `uint64`. `payloadReadOffset` is then used to
re-slice:

```go
chunkData := payloadChunk.Data[payloadReadOffset:]
```

A chunk whose leading uvarint is eleven `0xff` bytes panics:
`slice bounds out of range [-11:]`.

A negative offset also corrupts the completeness check just above it, since
`uint64(payloadReadOffset)` on a negative int is enormous.

## 2. The declared total is used as an allocation size, unbounded

```go
Data: make([]byte, self.txChunkSize),
```

`txChunkSize` is a `uint64` straight from the wire, with no upper bound. A peer
can declare a total large enough to exhaust memory, or on a 32-bit platform to
fail the make outright.

## 3. The write offset advances by the incoming length, not the bytes copied

```go
copy(self.txChunkPayload.Data[self.txChunkWriteOffset:], chunkData)
self.txChunkWriteOffset += len(chunkData)
payloadComplete := uint64(self.txChunkWriteOffset) == self.txChunkSize
```

`copy` is bounded by the destination, but the offset is not. A peer that sends
more bytes than it declared pushes `txChunkWriteOffset` past
`len(txChunkPayload.Data)`, so:

- `payloadComplete` compares past the declared total and is never true again
- the next chunk's `Data[self.txChunkWriteOffset:]` re-slice panics

In practice the first symptom wins. Declaring 8 bytes and then sending two
6-byte chunks does not panic and does not error: `nextTxPayload` loops back to
wait for a chunk that will never complete the payload, and blocks forever while
pinning the reassembly buffer. A test that exercises it hangs until the Go test
timeout rather than failing. That makes it a silent wedge of the circuit's read
path, not just a crash.

## Fix

Reject a non-positive uvarint count. Bound the declared total so it is exact as
an `int` on 32-bit platforms, matching what the channel frame decoder now does.
Reject a chunk that would overrun the declared total, and advance the write
offset by the bytes actually copied.

The representability bound is deliberately not a policy limit on payload size.
Choosing a real maximum belongs with the channel max-message-size work.

plorenz:
Fixes #1004, fixes #1006.

Three commits, in dependency order, each independently building and testing green.

**Update dependencies and regenerate the release notes.** Brings every direct
dependency current, superseding the dependabot group in #1012. That PR bumped
versions without running `go mod tidy`, and `acceptance`, `example` and
`example/influxdb-client-go` each carry
`replace github.com/openziti/sdk-golang/v2 => ../`, so a root-only bump leaves them
inconsistent and CI reads all three. pfxlog is held at v0.6.10, since v1.0.0 drops
`WithError`, `Tracef` and `ContextLogger` that `xgress` still uses; taking it is the
slog migration rather than a version bump. Release notes regenerated with
`ziti-ci update-sdk-release-notes`.

The channel move to v5.0.29 lands in this commit rather than later because its
frame reader classifies a short frame by magic before requiring a full frame's
bytes. On v5.0.10 a version-negotiation response came back as an unexpected EOF, so
the decoder work cannot be tested without it.

**Bounds-check the packet payload decoder (#1004).** `UnmarshallPacketPayload` read
its buffer on trust from its first line onwards. It is the packet producer for link
underlays, and on a datagram underlay `rxHello` uses the same read path as any other
frame, so it runs ahead of the bind handler. Every length is now validated;
`readU8ToBytesMap` compares in `uint64` rather than narrowing to `int`, which went
negative above `MaxInt64` and admitted the value before panicking; a delegated V2
frame has its declared lengths and each header entry checked against the datagram;
and magic is classified first, so a frame that sets bit 0 without being V2 reaches
the reader and returns the typed errors that drive version negotiation.

**Bound and grow chunked payload reassembly (#1006).** The reassembly total is
peer-declared and was trusted three ways. The buffer now grows as chunks arrive
rather than being allocated at the declared size, so a peer declaring 2GB and
sending 6 bytes costs 78KB. A non-positive `binary.Uvarint` count is rejected before
being used as an offset, and a chunk overrunning the declared total is refused;
previously that neither panicked nor errored, it blocked the read path indefinitely
while holding the buffer. Note the receiver enters this path on the flag in the
received payload, not on local config, so a peer reaches it even though nothing in
ziti enables chunking on the send side.

The `int32` bounds throughout are representability limits so values stay exact on
32-bit platforms, not policy limits on size.

Verified on amd64 and `GOARCH=386`, with a fuzz target over the decoder committed
alongside the input that found the `readU8ToBytesMap` defect.


xgress/xgress.go
```diff
@@ -1171,21 +1204,85 @@ func (self *Xgress) addGoroutineIfRelated(buf *bytes.Buffer, xgressRelated bool,
...
+
+// UnmarshallPacketPayload decodes one datagram into a message. buf is peer-supplied and
+// arbitrary; every malformed input yields an error rather than a panic. Callers get a message
+// that borrows buf, so buf must not be reused until the message is done with.
 func UnmarshallPacketPayload(buf []byte) (*channel.Message, error) {
+	if len(buf) < 1 {
+		return nil, errors.New("payload is empty, no flags byte")
+	}
 	flagsField := buf[0]
 	if flagsField&1 != 0 {
```
andrewpmartinez, line 1281: A datagram with bit 0 set but no V2 magic now gets a length error rather than a bad magic error. Is it worth a magic check ahead of the length validation, so the message says what actually went wrong?

xgress/xgress.go
```diff
@@ -594,6 +595,26 @@ func (self *Xgress) processTxPayload(payload *Payload) ([]byte, map[uint8][]byte
...
+// parseChunkedPayloadSize reads the reassembly total leading a chunked payload, returning it with
+// the number of bytes it occupied. The total is peer-supplied and becomes an allocation size, so
+// it is validated before any caller can act on it. Kept separate from the reassembly loop so the
+// bound can be tested at its edges without a test having to reach the allocation it guards.
+func parseChunkedPayloadSize(data []byte) (uint64, int, error) {
+	size, read := binary.Uvarint(data)
+	if read < 1 {
+		return 0, 0, errors.New("chunked payload size is truncated or overflows a uint64")
+	}
+	if size > maxChunkedPayloadSize {
```
andrewpmartinez, line 615: MaxInt32 still lets a peer make the receiver allocate 2GB for a payload it has sent a few bytes of. Do you want a policy limit here rather than wait for the channel max message size work?

andrewpmartinez (approved):
Two comments, neither are terible, but allocating 2GB from a peer supplied value might be a problem.

xgress/xgress.go
plorenz: fixed

xgress/xgress.go
plorenz: converted the chunk assembly code to pre-allocate a max of of 256KB and to grow the buffer via append once it gets past that, only once data actually arrives.

PR #1008, opened 2026-08-25
[Backport-1.8] Refuse dials for unknown edge conn ids

### LINKED ISSUE #1007: [Backport-1.8] Dial for an unknown edge conn id is dropped silently, costing the client its full connect timeout
Backport of #999 to `release-v1.8.x`.

## Why

`ConnMuxImpl.HandleReceive` logs and drops a dial addressed to a conn id it has no sink for. A hosted
listener can close while its terminator is still selectable at the controller, so the router forwards
dials for it whenever a listener closes with dials in flight. The router then waits for a reply under a
route timeout longer than the dialing client's connect timeout, so the silence reaches the client as a
connect timeout instead of a fast, retriable failure.

Fixed on the development line by c1a6236d80889bea9febc9082915ac00e5691041 (#1000), which replies
`DialFailed` so the router learns the terminator is unusable and can fail over.

## Why it matters on this line

`openziti/ziti` `release-v2.0.x` is pinned to the v1 line and is on `channel/v4`, so it cannot take the
fix from `main`: `main` is now `sdk-golang/v2` on `channel/v5`, and v1.9.0 also moved to `channel/v5`.
`release-v1.8.x` is the newest v1 line still on `channel/v4`, which makes it the only place this fix can
reach that consumer.

It also shows up as a CI flake there, in `Test_HSRotatingDataflow`, which rotates hosted listeners and so
reproduces the close-with-dials-in-flight window directly.

## Porting notes

The change is small and additive, and the surrounding code is unchanged on this line, but it needs the
v5 to v4 translation:

- `ch.GetSenders().(SdkChannel)` does not exist on `channel/v4`. The equivalent is already used in this
  same file by `returnInspectResponse`: assert `ch.(channel.MultiChannel)`, then
  `GetUnderlayHandler().(SdkChannel)`.
- The test's channel double declares `GetSenders`, which is not part of the v4 interface, and calls
  `mux.SetSdkChannel`, which does not exist here. Neither is needed: on v4 the double simply must not
  satisfy `channel.MultiChannel`, so the sender lookup falls back to the channel itself.

plorenz:
Fixes #1007. Backport of #1000 to `release-v1.8.x`.

`ConnMuxImpl.HandleReceive` logged and dropped a dial addressed to a conn id it has no sink for. A hosted
listener can close while its terminator is still selectable at the controller, so the router forwards dials
for it whenever a listener closes with dials in flight. The router then waits for a reply under a route
timeout longer than the dialing client's connect timeout, so the silence reached the client as a connect
timeout instead of a fast, retriable failure. It now replies `DialFailed`, so the router can fail over to
another terminator.

### Ported rather than cherry-picked

`main` is `sdk-golang/v2` on `channel/v5`; this line is on `channel/v4`, so one call needed translating.
The surrounding code is otherwise unchanged here and the patch applied cleanly.

| v5 (`main`) | v4 (this branch) |
|---|---|
| `ch.GetSenders().(SdkChannel)` | `ch.(channel.MultiChannel)` then `GetUnderlayHandler().(SdkChannel)` |

That is not a new idiom: `returnInspectResponse` in this same file already reaches the control sender
exactly that way. The test's channel double drops its `GetSenders` method and the `SetSdkChannel` call for
the same reason. Neither exists on v4, and neither is needed: the double only has to not satisfy
`channel.MultiChannel` for the refusal to fall back to sending on the channel itself, which is where the
test captures it. Nothing else differs from the original commit.

### Why this line

`openziti/ziti` `release-v2.0.x` is pinned to the v1 line and is on `channel/v4`. It cannot take the fix
from `main`, and v1.9.0 also moved to `channel/v5`, so `release-v1.8.x` is the newest v1 line that can
carry it. It surfaces there as a CI flake in `Test_HSRotatingDataflow`, which rotates hosted listeners and
so reproduces the close-with-dials-in-flight window directly.

### Verification

Build and vet clean, full suite passes. The new test was checked against its own regression: with the
`ContentTypeDial` branch removed it fails with "no reply sent for a dial addressed to an unknown conn id",
and passes with it.

### DIFF
--- a/ziti/edge/msg_mux.go
+++ b/ziti/edge/msg_mux.go
@@ -374,9 +374,37 @@ func (mux *ConnMuxImpl[T]) HandleReceive(msg *channel.Message, ch channel.Channe
 		mux.handlePayloadWithNoSink(msg, ch)
 	} else if msg.ContentType == ContentTypeStateClosed {
 		// ignore, as conn is already closed
+	} else if msg.ContentType == ContentTypeDial {
+		go mux.handleDialWithNoSink(connId, msg, ch)
 	} else {
 		pfxlog.Logger().WithField("connId", connId).WithField("contentType", msg.ContentType).
 			Info("unable to dispatch msg received for unknown edge conn id")
+	}
+}
+
+// handleDialWithNoSink refuses a dial addressed to a conn this mux does not have.
+//
+// The router keeps forwarding dials for a hosted conn until its terminator is removed,
+// which happens after the conn is gone from here, so this arrives whenever a listener
+// closes while dials are in flight. The router waits for a reply under a route timeout
+// longer than the dialing client's connect timeout, so dropping it silently costs the
+// client its whole timeout; refusing it lets the router fail over to another terminator.
+func (mux *ConnMuxImpl[T]) handleDialWithNoSink(connId uint32, msg *channel.Message, ch channel.Channel) {
+	pfxlog.Logger().WithField("connId", int(connId)).
+		Info("refusing dial for unknown edge conn id")
+
+	var sender channel.Sender = ch
+	if mc, ok := ch.(channel.MultiChannel); ok {
+		if sdkChan, ok := mc.GetUnderlayHandler().(SdkChannel); ok {
+			sender = sdkChan.GetControlSender()
+		}
+	}
+
+	reply := NewDialFailedMsg(connId, fmt.Sprintf("no listener for conn id [%v]", connId))
+	reply.ReplyTo(msg)
+	if err := reply.WithPriority(channel.Highest).WithTimeout(5 * time.Second).Send(sender); err != nil {
+		pfxlog.Logger().WithFields(GetLoggerFields(msg)).WithError(err).
+			Error("failed to refuse dial for unknown edge conn id")
 	}
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Release notes 1.8.2
+
+## Issues Fixed and Dependency Updates
+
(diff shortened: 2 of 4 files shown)


ziti/edge/msg_mux.go
```diff
@@ -374,12 +374,40 @@ func (mux *ConnMuxImpl[T]) HandleReceive(msg *channel.Message, ch channel.Channe
...
+	var sender channel.Sender = ch
+	if mc, ok := ch.(channel.MultiChannel); ok {
+		if sdkChan, ok := mc.GetUnderlayHandler().(SdkChannel); ok {
+			sender = sdkChan.GetControlSender()
+		}
+	}
+
+	reply := NewDialFailedMsg(connId, fmt.Sprintf("no listener for conn id [%v]", connId))
+	reply.ReplyTo(msg)
+	if err := reply.WithTimeout(5 * time.Second).Send(sender); err != nil {
```
andrewpmartinez: Should this send match the other dial failure replies?

- Both sends in `newChildConnection` use `WithPriority(channel.Highest)` and `SendAndWaitForWire`.
- On a channel that is not multi-underlay the sender here is the channel itself, so a default-priority refusal queues behind pending data, which is the delay the fix is trying to avoid.

andrewpmartinez (approved):

ziti/edge/msg_mux.go
plorenz: SendAndWaitForWire doesn't make a difference b/c nothing is waiting the send here. We just care that it queued. If there's a failure, the channel is getting torn down, this send becomes irrelevant. 
Added WithPriority, but in most cases we should be on the control channel already and it shouldn't matter. WithPriority is removed in channel v5

PR #1022, opened 2026-09-16
Close a legacy edge conn on a StateClosed that arrives after the FIN

### LINKED ISSUE #1018: Legacy edge conn ignores a StateClosed that arrives after a FIN
## Summary

A legacy (non flow-control) edge conn stops dequeuing messages once it has read a FIN: `Read`
returns `io.EOF` immediately whenever `readFIN` is set, without touching the sequencer. A
`StateClosed` message is only acted on when `Read` dequeues it, so a `StateClosed` that arrives
after the FIN is never processed. The conn stays open from the SDK's point of view: `IsClosed()`
is false, the write side is still accepted locally, and the mux sink is never removed until the
application calls `Close` itself.

`AcceptMessage` only marks `sentFIN` and queues the `StateClosed`; it does not close.

## When it happens

Any time the router sends a FIN followed by a `StateClosed` on the same conn. The router does this
whenever a terminator half-closes and the circuit is later torn down, and, since
openziti/ziti#4338 and #4394, whenever an xgress terminator's EOF is relayed to a legacy client:
the initiator xgress's tx goroutine sends the FIN while the unroute path sends `StateClosed`, both
on the default sender, so the order is a race. When the FIN wins, the client never learns the
conn was closed. This is what makes `Test_PostureRevalidation_HostTerminator_OnPostureData_OIDC`
flake on the ziti 2.0.x backport PR (openziti/ziti#4341): revocation works, but the client conn
never reports closed.

## Fix

Close the conn once both the FIN has been read and `StateClosed` has been received, whichever
comes last:

- `AcceptMessage` records that `StateClosed` was received, then closes immediately if the FIN has
  already been read.
- `Read` checks for a received `StateClosed` at the point it returns `io.EOF` because of the FIN,
  and closes if one is pending.

Each side stores its own flag before loading the other's, so with sequentially consistent atomics
at least one side always observes both, and `close` is already idempotent. A `StateClosed` that
arrives before the FIN still flows through the queue in order, as today. Flow-control (xgress)
conns have their own `StateClosed` handling and are unchanged.

Affects the v2 module's `edgeConnLegacy` and the v1 line's `edgeConn`.

plorenz:
Fixes #1018.

A legacy edge conn stops draining its queue once it has read a FIN: `Read` returns `io.EOF` from then on without dequeuing anything. A `StateClosed` is only acted on when `Read` dequeues it, so one that arrives after the FIN is never processed and the conn stays open from the SDK's point of view until the application calls `Close`. The router produces exactly that sequence whenever a half-closed circuit is later torn down, and since openziti/ziti#4338 whenever an xgress terminator's EOF is relayed to a legacy client; the FIN and the StateClosed race, and when the FIN wins the client never learns the conn closed.

The conn now closes once both "read side reached EOF" and "StateClosed received" have happened, whichever comes second. The two events are bits of one `AtomicBitSet`: each side sets its bit with `SetAndGetPrevious` and closes if the other bit was already present, so a single read-modify-write settles the race. A StateClosed that arrives before any FIN still flows through the queue in order. Flow-control conns are unchanged.

`SetAndGetPrevious` is openziti/foundation#500; go.mod pins its commit until v2.0.102 is tagged. The release-notes tool did not pick up this branch, so the changelog entry is hand-written in its format.


dovholuknf (approved):
code seems fine to me

